PR #5, opened 2016-10-27
Add contribution section to README

### LINKED ISSUE #2: Add a "Contributing" section to the README
We need to have a section the README that explains that we're looking for contributors and how those wonderful souls can get involved.

smudge:
/domain @samandmoore @jmileham 

Seems like this would resolve #2. Not sure what exactly we had in mind so I figured I'd throw something out there and see if there are any objections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ A name for a customer identifier that is meaningful in your application, typical
 
 ## How to Contribute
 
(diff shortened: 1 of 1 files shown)


jmileham:
<< domain lgtm! Thanks @smudge. We should get @samandmoore's LG as well and then roll out the section to all the repos?

samandmoore:
lgtm! exactly what we had in mind. i think we should use this same blurb for all the TT repos.

PR #8, opened 2016-10-31
Allow bulk assignments to decide variant

rynonl:
/domain @aburgel 
/platform @jmileham 

This allows for `variant` to be left blank on bulk assignments. If it is blank, the bulk_assignment will run the variant calculator and assign visitors as if they'd actually experienced the split. It also will not override existing assignments since these would have been deterministically decided already, or overridden via the chrome plugin which we want to honor.


rynonl:
Closing per offline discussion, needs more thought about long term solutions for eagerly assigned splits.

PR #10, opened 2016-11-02
Upgrade SAML libs and make the signout button work

jmileham:
The signout route previously worked differently with password auth

/domain @samandmoore @rynonl @aburgel


samandmoore:
<<domainlgtm -- good catch on the broken link/button

PR #16, opened 2016-12-12
Batch decisions

jmileham:

There are a couple other things in here - there are dead templates related to long dead routes I removed because I saw them. Also, a test was relying on created_at return sequence of assignments in the visitor. The performance tradeoff should be tiny to make those deterministic, even without an index on created_at, because a visitor has a very small number of assignments to return.

### DIFF
--- a/app/models/decision.rb
+++ b/app/models/decision.rb
@@ -32,9 +32,9 @@ class Decision
   end
 
   def reassign_existing_assignments
-    transaction do
-      bulk_assignment.save!
-      BulkReassignment.create!(assignments: existing_assignments, bulk_assignment: bulk_assignment)
+    bulk_assignment.save!
+    existing_assignments.find_in_batches do |subset|
+      BulkReassignment.create!(assignments: subset, bulk_assignment: bulk_assignment)
     end
   end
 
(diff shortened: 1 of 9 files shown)


app/models/decision.rb
```diff
@@ -32,9 +32,9 @@ def weight_split_fully_to_variant
   end
 
   def reassign_existing_assignments
-    transaction do
-      bulk_assignment.save!
-      BulkReassignment.create!(assignments: existing_assignments, bulk_assignment: bulk_assignment)
+    bulk_assignment.save!
+    existing_assignments.find_in_batches do |subset|
```
reustmd: any idea how much faster this will be, since it's still wrapped in an outer transaction?

in mysql-land i wouldn't expect a huge performance win, since the transaction log still has to grow so dang much.

jmileham: This isn't to improve performance in web context - it'll possibly actually be slower in web, but the web thread will time out. This is actually to be run in a background context without a wrapping transaction, and then we'll get reasonable throughput and no blocking except for the batch currently under consideration for as long as that txn takes to commit, presumably well sub-second.

aburgel:
<< domainlgtm

PR #20, opened 2016-12-20
Add support for DelayedJob

samandmoore:
/domain @rynonl @jmileham 
/cc @reustmd @dschaub @while1malloc0 

This changeset includes support for delayed job. It's not being utilized yet, but I thought it would be a good idea to land this as it's own PR to keep things small.

I'm open to expanding on this PR to include a feature that uses DJ.


jmileham:
<< domain lgtm

PR #22, opened 2016-12-20
rely on external supervisor for puma

samandmoore:
/domain @jmileham 

since we've got a Procfile now and we're endorsing foreman-backed deployment processes that include their own process supervisor, we don't need nor want to daemonize our puma process nor track its PID.


jmileham:
<< domain lgtm

PR #26, opened 2016-12-22
move the split decision execution into an activejob

samandmoore:
this operation can take a long time to execute on a split that has lots
of assignments to reassign. this change pushes the execution of
reassignment into a background job. this change also trades off
guarantees that we only create a single BulkAssignment for the Decision.
in the non-optimal case, the job will die while executing the
reassignments in batches. when that happens, the job will re-run and
pick up the reassignment process where it left off with the one
unfortunate side effect of creating another BulkAssignment.

/domain @jmileham @rynonl 

Note: this PR also introduces ActiveJob over using `split.delay.create_decision!(...)`. I may have gotten the joint wrong, but it feels okay.

### DIFF
--- a/app/controllers/admin/decisions_controller.rb
+++ b/app/controllers/admin/decisions_controller.rb
@@ -6,19 +6,15 @@ class Admin::DecisionsController < AuthenticatedAdminController
 
   def create
     split = Split.find params[:split_id]
-    decision = split.build_decision(create_params)
-    CreateDecisionJob.perform_later(split, create_params)
-    flash[:success] = "Queued decision to reassign #{decision.count} visitors to #{decision.variant}"
+    affected_assignments = split.assignments.where.not(variant: target_variant)
+    CreateDecisionJob.perform_later(split, variant: target_variant, admin: current_admin)
+    flash[:success] = "Queued decision to reassign #{affected_assignments.count} visitors to #{target_variant}"
     redirect_to admin_split_path(split)
   end
 
   private
 
-  def create_params
-    create_form_params.merge(admin: current_admin)
-  end
-
-  def create_form_params
-    params.require(:decision).permit(:variant)
+  def target_variant
+    params.require(:decision).fetch(:variant)
   end
 end
--- a/app/controllers/admin/decisions_controller.rb
+++ b/app/controllers/admin/decisions_controller.rb
@@ -5,18 +5,17 @@ class Admin::DecisionsController < AuthenticatedAdminController
   end
 
   def create
-    @split = Split.find params[:split_id]
-    Split.transaction do # FIXME: This makes sure this controller either succeeds or fails atomically until we can async the work
-      @decision = @split.create_decision!(create_params)
-    end
-    flash[:success] = "Reassigned #{@decision.count} visitors to #{@decision.variant}"
-    redirect_to admin_split_path(@split)
+    split = Split.find params[:split_id]
+    decision = split.build_decision(create_params)
+    CreateDecisionJob.perform_later(split, create_params)
+    flash[:success] = "Queued decision to reassign #{decision.count} visitors to #{decision.variant}"
+    redirect_to admin_split_path(split)
   end
 
   private
 
   def create_params
-    create_form_params.merge(admin: current_admin, split: @split)
+    create_form_params.merge(admin: current_admin)
   end
 
   def create_form_params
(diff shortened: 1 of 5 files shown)


app/controllers/admin/decisions_controller.rb
```diff
@@ -5,18 +5,17 @@ def new
...
-      @decision = @split.create_decision!(create_params)
-    end
-    flash[:success] = "Reassigned #{@decision.count} visitors to #{@decision.variant}"
-    redirect_to admin_split_path(@split)
+    split = Split.find params[:split_id]
+    decision = split.build_decision(create_params)
+    CreateDecisionJob.perform_later(split, create_params)
+    flash[:success] = "Queued decision to reassign #{decision.count} visitors to #{decision.variant}"
+    redirect_to admin_split_path(split)
   end
```
samandmoore: in this controller action we have to build the decision in order to provide the same level of feedback to the admin performing the decision. we end up throwing that decision away because we can't serialize an ActiveModel in ActiveJob. At least, it wasn't immediately obvious how to accomplish that.

app/controllers/admin/decisions_controller.rb
```diff
@@ -5,18 +5,17 @@ def new
...
 
   private
 
   def create_params
-    create_form_params.merge(admin: current_admin, split: @split)
+    create_form_params.merge(admin: current_admin)
```
samandmoore: this `split:` option was redundant with the `Split#build_decision` method's internals: https://github.com/Betterment/test_track/blob/master/app/models/split.rb#L64

app/models/decision.rb
```diff
@@ -9,8 +9,6 @@ class Decision
 
   validate :variant_belongs_to_split
 
-  delegate :transaction, to: BulkAssignment
```
samandmoore: this doesn't appear to be used.

spec/features/admin_split_decision_create_spec.rb
```diff
@@ -36,6 +36,12 @@
...
     expect(split_page).to be_displayed
+    expect(split_page).to have_content "Queued decision"
+
+    Delayed::Worker.new.work_off
+
+    split_page.load split_id: split.id
+    expect(split_page).to be_displayed
 
     expect(Assignment.where(variant: "hammer_time").count).to eq 0
     expect(Assignment.where(variant: "touch_this").count).to eq 5
```
samandmoore: We're using the `:delayed_job` adapter in all environments. I think that's saner than the rails default of using a test adapter which tracks enqueued jobs and can potentially run them inline. I also am unsure of how that test adapter behaves in feature specs. This approach felt clean to me: enqueue jobs same as ever, then choose to work them off in order to observe their side effects.

rynonl:
Awesome

<< domainlgtm

spec/features/admin_split_decision_create_spec.rb
jmileham: Cool, this works well for the rollback isolation strategy. We'd just want to be sure to purge the delayed_jobs table between specs if using a deletion/truncation strategy. Perhaps that'd be automatic too.

jmileham: I guess this is using truncation strategy being a feature spec, and yeah, it works by default because that's how DatabaseCleaner works unless you exclude the table, eh? :)

jmileham:
double lg. this looks good.

app/controllers/admin/decisions_controller.rb
```diff
@@ -5,21 +5,22 @@ def new
...
-    @split = Split.find params[:split_id]
-    Split.transaction do # FIXME: This makes sure this controller either succeeds or fails atomically until we can async the work
-      @decision = @split.create_decision!(create_params)
-    end
-    flash[:success] = "Reassigned #{@decision.count} visitors to #{@decision.variant}"
-    redirect_to admin_split_path(@split)
+    CreateDecisionJob.perform_later(split, variant: target_variant, admin: current_admin)
+    flash[:success] = "Queued decision to reassign #{affected_assignments_count} visitors to #{target_variant}"
+    redirect_to admin_split_path(split)
   end
```
samandmoore: @jmileham rubocop wouldn't leave me alone about the ABC size of this method until i moved all but one of the assignments into a method... so i did that. we lose the "grab the count before we delay the work" thing though. should we just disable the cop on this method?

PR #27, opened 2017-03-02
Fix case sensitivity bug in SAML auth

alianam:
/domain @samandmoore

### DIFF
--- a/spec/controllers/api/v1/visitors_controller_spec.rb
+++ b/spec/controllers/api/v1/visitors_controller_spec.rb
@@ -36,18 +36,32 @@ RSpec.describe Api::V1::VisitorsController, type: :controller do
 
         expect(response).to have_http_status :ok
         expect(response_json["id"]).to eq(visitor.id)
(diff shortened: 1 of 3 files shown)


spec/controllers/api/v1/visitors_controller_spec.rb
```diff
@@ -36,18 +36,32 @@
...
+        expect(response_json["assignments"].length).to eq 3
+        expect(response_json["assignments"]).to include(
+          hash_including(
+            "split_name" => "one",
+            "variant" => "control",
+            "unsynced" => false,
+            "context" => "context_a"
+          )
+        )
+
```
samandmoore: nitpick here: can we add a line break like these between each of these `expect` sections or none of them? :)

samandmoore:
<<domain

just a tiny comment about spacing, but

domainlgtm!

great work on this! that's a sneaky bug.

jmileham:
Removed the extra newline. Good eye. Mergin'. thanks @alianam!

PR #28, opened 2017-03-08
Delay mass assignments

dschaub:
/domain @samandmoore @while1malloc0 @agirlnamedsophia

### DIFF
--- a/app/models/bulk_assignment_creation.rb
+++ b/app/models/bulk_assignment_creation.rb
@@ -25,6 +25,10 @@ class BulkAssignmentCreation
     true
   end
 
+  def save!
+    save || raise(ActiveRecord::RecordNotValid)
+  end
+
   def self.create(params)
     new(params).tap(&:save)
   end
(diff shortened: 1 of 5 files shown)


while1malloc0:
<< domain

app/models/bulk_assignment_creation.rb
```diff
@@ -25,6 +25,10 @@ def save
     true
   end
 
+  def save!
+    save || raise(ActiveRecord::RecordNotValid)
```
while1malloc0: Should this be `raise ActiveRecord::RecordInvalid, self`?

while1malloc0:
<< domain lgtm

app/models/bulk_assignment_creation.rb
samandmoore: (i think so)

samandmoore:
<<domainlgtm

while1malloc0:
Oh right, I'm not on test track core. Derp 😄

PR #30, opened 2017-03-16
Upgrade style closet and spec helpers

dschaub:
### Summary

Updating style closet and spec helpers ahead of updating the split details page design. Spec helpers just came with some updated capybara helpers and metaspecs for view security.

Everything looks pretty much exactly the same after upgrading style closet.


samandmoore:
<<domainlgtm 

we're going to do a PR that includes a once-over of the style-closet usages in here? i don't remember what this app is doing w/r/t typography, etc, but there have been many improvements to SC that we probably wanna integrate.

also, props to SC for not having a crapton of breaking changes since october.

dschaub:
yeah, once this is in we'll start using the new stuff in SC on the details page.

PR #31, opened 2017-03-17
Add details to split

agirlnamedsophia:
### Summary

> This enables an admin to see and edit context and hypothesis metadata about what splits the user is experiencing

### Other Information

> design updates forthcoming in a separate PR to more completely update the UI with style-closet

hypothesis


app/views/admin/split_contexts/edit.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="form-container sc-m-t--l"%>
+  <%= simple_form_for @split, url: admin_split_context_path, method: :put do |f| %>
+    <%= f.error :base %>
+    <%= f.input :hypothesis, hint: "What do we expect to change with the treatment condition, and why?" %>
+    <%= f.input :context, hint: "When do we assign users to this split? Is the split population constrained to a particular type of user or action? Assignments should happen as \"close\" to the varied experience as possible, so that only relevant users are included in the analysis." %>
```
rynonl: I wonder if we want to double up on our usage of `context` in TT...it's already an attribute on `assignments`.  Perhaps `population` or even just a more descriptive `assignment_context`?

samandmoore: i left a similar comment below but without all this wonderful reasoning. i was suggesting "description" but it sounds like this is trying to be more specific than that. is there going to be a general description field for a split, or would it be fine to call this description and just give a really good prompt for what should be put in this field?

app/assets/stylesheets/_card.scss
```diff
@@ -0,0 +1,53 @@
...
+      float: none;
+      width: auto;
+      padding-bottom: 20px;
+    }
+  }
+
+  .action-form {
+    padding: $card-padding;
+  }
+}
```
samandmoore: we should take a whack at using SUIT for all of these things and rejigger this SCSS file structure to use 7+1 while we're in here replacing all the SCSS. also, we should rip out bootstrap :)

samandmoore: ... in another PR ;)

agirlnamedsophia: already started! I didn't want to overwhelm the scope of this PR.

app/views/admin/splits/_details.html.erb
```diff
@@ -0,0 +1,35 @@
...
+            <td>&nbsp;</td>
+            <td>
+                <div class="split-decision">
+                    <span class="flag">🏁</span>
+                    <%= link_to "Decide Split", new_admin_split_decision_path(split) %>
+                </div>
+            </td>
+        </tr>
+    </table>
+</article>
```
samandmoore: this file seems like it has multiple different indent styles. i think we're doing 2 spaces in all html

agirlnamedsophia: Ah you're right, html files here have 4 spaces throughout. The "ArticleHeader" addition was my own, which follows 2 space-spacing. 

Does it make sense to tackle the discrepancy in spacing throughout in this PR or in the layout-revamp-change? Also, should we follow 2 or 4 spaces? 

(Of course ensuring this file at least is consistent for this PR).

samandmoore: 2 spaces for everything except scss is our usual convention. i would just use 2 spaces for the files you added / changed / extracted here. and then we can follow that each time we change something going forward. maybe we will do a separate PR that just re-indents everything later.

app/views/admin/split_contexts/edit.html.erb
samandmoore: another idea: `"background"`?

samandmoore: although, `"description"` is really starting to seem right. are we going to add another field called "description" that will be different or is that what this really is?

samandmoore:
<<domain tafn

app/views/admin/split_contexts/edit.html.erb
agirlnamedsophia: The designs are calling it `Assignment Context` so maybe it makes more sense to call the column that as well.

agirlnamedsophia:
bump

agirlnamedsophia:
nanda?

db/migrate/20170317155628_add_details_to_splits.rb
```diff
@@ -0,0 +1,8 @@
+class AddDetailsToSplits < ActiveRecord::Migration
+  def change
+    add_column :splits, :hypothesis, :text, null: true
+    add_column :splits, :assignment_criteria, :text, null: true
+    add_column :splits, :description, :text, null: true
+    add_column :splits, :squad_owner, :string, null: true
```
samandmoore: It seems to me that, broadly, engineering teams would be capable of identifying the owner of a given split test, so the idea of a split having an owner is perfectly reasonable. I think I would just drop the "Squad" part from the name since it's not really something that means anything to the TestTrack product. And if owner is ambiguous in the UI, then we should feel comfortable adding another example text hint that clarifies the kinds of values that would be reasonable for this free text field.

app/models/split.rb
```diff
@@ -19,6 +19,10 @@ class Split < ActiveRecord::Base
 
   scope :active, -> { where(finished_at: nil) }
 
+  def has_details?
+    %w(hypothesis assignment_criteria description squad_owner).any? { |attr| send(attr).present? }
```
samandmoore: minor nit: we can use `public_send` here because these are all public attributes.

db/migrate/20170317155628_add_details_to_splits.rb
samandmoore: Also, TestTrack already has a large glossary of terms surrounding it, so i'm not super psyched about adding another term -- "Squad" -- to the list. I'm fine with being overruled and we just add the idea of squads as a unit that owns splits to the readme somewhere. We just don't want to lose focus of the fact that this is a product that we should strive to keep as simple as humanly possible. ❤️

samandmoore:
minor comment about one line of code and a philosophical point about the naming of that one column, but domainlgtm

let's just sync up on the name of that column. 

very excited to start collecting these details in the app properly so that we can build a kick-butt analytics tool on the backside of this database.

agirlnamedsophia:
bump

app/controllers/admin/split_details_controller.rb
```diff
@@ -0,0 +1,21 @@
+class Admin::SplitDetailsController < AuthenticatedAdminController
+  def edit
+    @split = Split.find params[:split_id]
+  end
+
+  def update
+    @split = Split.find params[:split_id]
+    if @split.update!(split_context_params)
```
samandmoore: Missed this before but we should be using update not update! here otherwise we will never hit the else path. 

Is there no feature spec for this feature? Dun dun dun. I think I missed that before as  well :(

agirlnamedsophia:
bump

samandmoore:
domainlgtm part 2! nice.

PR #32, opened 2017-03-20
Updates stylecloset

agirlnamedsophia:

* This updates test track's style closet so we can implement `layout` styles.


samandmoore:
<<domainlgtm

PR #34, opened 2017-03-23
Add table and edit form for variant metadata

dschaub:
### Summary

Adding the data model and initial form for adding human-readable metadata to variants. I'd like to wait for @agirlnamedsophia to land her style changes before this gets merged, but I think it's ready to look at.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,4 +7,4 @@ indent_style = space
 indent_size = 2
 
 [*.erb]
-indent_size = 4
+indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,6 +5,3 @@ end_of_line = lf
 insert_final_newline = true
 indent_style = space
 indent_size = 2
-
-[*.erb]
-indent_size = 2
--- a/app/models/variant_detail.rb
+++ b/app/models/variant_detail.rb
@@ -2,6 +2,7 @@ class VariantDetail < ActiveRecord::Base
   belongs_to :split
 
   validate :variant_must_exist, on: :create
+  validates :display_name, :description, presence: true
 
   def display_name
     super || variant
--- a/app/models/variant_detail.rb
+++ b/app/models/variant_detail.rb
@@ -1,14 +1,15 @@
-class VariantDetail
-  attr_reader :split, :variant, :weight, :assignment_count
+class VariantDetail < ActiveRecord::Base
+  belongs_to :split
 
-  def initialize(split, variant)
-    @split = split
-    @variant = variant
-    @weight = split.variant_weight(variant)
-    @assignment_count = split.assignment_count_for_variant(variant)
+  validate :variant_must_exist, on: :create
+
+  def display_name
+    super || variant
   end
 
-  def retirable?
-    weight == 0 && assignment_count > 0
+  private
+
+  def variant_must_exist
+    errors.add(:base, "Variant does not exist: #{variant}") unless split.has_variant?(variant)
   end
 end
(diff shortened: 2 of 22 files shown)


db/migrate/20170316204717_create_variant_details.rb
```diff
@@ -0,0 +1,15 @@
+class CreateVariantDetails < ActiveRecord::Migration
+  def change
+    create_table :variant_details, id: :uuid do |t|
+      t.uuid :split_id, null: false, index: true
+      t.string :variant, null: false
+      t.string :display_name
+      t.text :description
```
samandmoore: Should these be null: false and required through validations? We don't really get any value out of having one of these rows without these two columns populated. And we're creating these lazily, so i think it makes sense to be non nullable.

app/views/admin/splits/_variants.html.erb
```diff
@@ -28,6 +29,7 @@
                         EOF
               <% end %>
             </td>
+            <td><%= link_to 'edit', edit_admin_split_variant_detail_path(@split, variant_detail.variant) %></td>
```
agirlnamedsophia: do these `@split`s need to be ivars here?

spec/factories/variant_details.rb
```diff
@@ -0,0 +1,8 @@
+FactoryGirl.define do
+  factory :variant_detail do
+    split
+    variant { split.variants.first }
+    display_name 'Great variant'
+    description 'Really, everyone loves it'
```
agirlnamedsophia: ❤️

.editorconfig
```diff
@@ -7,4 +7,4 @@ indent_style = space
 indent_size = 2
 
 [*.erb]
-indent_size = 4
+indent_size = 2
```
samandmoore: Also, if we're going to change this then just delete this because the above line says the same thing?

dschaub: yeah I can just kill it, I decided to keep the change in here because it was screwing with my spacing while writing the views.

app/views/admin/splits/_variants.html.erb
dschaub: nope!

dschaub:
bump

app/models/variant_detail.rb
```diff
@@ -1,14 +1,16 @@
...
-
-  def initialize(split, variant)
-    @split = split
-    @variant = variant
-    @weight = split.variant_weight(variant)
-    @assignment_count = split.assignment_count_for_variant(variant)
+class VariantDetail < ActiveRecord::Base
+  belongs_to :split
+
+  validate :variant_must_exist, on: :create
```
samandmoore: Shouldn't this apply all the time?

spec/features/admin_variant_details_edit_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  it 'allows admins to edit variant details' do
+    split_page.load split_id: split.id
+    expect(split_page).to be_displayed
+
+    expect(split_page.variants.count).to eq 1
+    split_page.variants.first.edit_link.click
+
+    expect(variant_page).to be_displayed
+
+    variant_page.form.display_name.set 'Variant name'
```
agirlnamedsophia: do we care about validating presence of these attributes via testing an empty form and submitting?

app/models/variant_detail.rb
samandmoore: Or, if we're really trying to optimize when this runs, it would apply on change of variant. But I'd be fine with just saying it applies always.

samandmoore:
<<domain lgtm

app/models/variant_detail.rb
dschaub: yeah I guess it can always run, I just didn't think there was a reason to prevent a variant that used to exist from having its metadata edited. but like we talked about before, that would only happen if someone was breaking the rules.

PR #35, opened 2017-03-25
Form support for new designs

agirlnamedsophia:
Summary of Changes
* new javascript files and their related form helpers to support new designs. waiting for the next PR to implement these forms with full deprecation of bootstrap
* added datetime helpers to support new designs
* added layout helper methods to support new layout styles (to be implemented in the next PR)

TBD
* need to determine if the `vendor_deps` rake task should be updated to support importing javascript files (or perhaps _only_ the javascript files we need) 


dschaub:
<< domain lgtm

PR #37, opened 2017-03-30
ByeByeBootstrap

agirlnamedsophia:
Summary
* removes bootstrap entirely; in so doing, needed to replace flash and icon functionality we were still getting, which means we needed a new unlicensed font
* also removed the form functionality bootstrap gave us and replaced it with our homegrown inputs
* in order to use the style closet inputs, needed to adjust a few site prism pages and their specs

/domain @asullivan210 { for scss }


config/initializers/simple_form.rb
```diff
@@ -142,4 +123,169 @@
...
+    radio_buttons: :radio_buttons,
+    radio_button_table_cell: :radio_button_table_cell,
+    descriptive_radio_buttons: :descriptive_radio_buttons,
+    button_group: :button_group,
+    check_boxes: :check_boxes,
+    boolean: :boolean,
+    datepicker: :datepicker,
+    percent: :percent,
+    date_string: :date_string,
+  }
```
asullivan210: I think some of these aren't needed and can be taken out

app/assets/stylesheets/components/_flash.scss
```diff
@@ -0,0 +1,10 @@
+.sc-flash {
+  .close {
+    margin-bottom: 1rem;
+  }
```
asullivan210: this doesn't JustWork with style closet styles?

agirlnamedsophia: No because of the new icon font, it's a different size :(

app/assets/stylesheets/components/_inputs.scss
```diff
@@ -1,3 +1,30 @@
...
+  input[type="checkbox"] + label.regular:before {
+    @include icomoon;
+    content: $icon-checkmark;
+    font-size: 107.5%;
+    padding: 0.04em 0 0 0.04em;
+  }
+}
+
+.input-radio {
+  padding-top: 1rem;
```
asullivan210: what top padding already exists on radio inputs with style closet? been trying to get rid of global top padding on inputs in retail because it isn't always wanted so I'm wondering if those styles can be more specific

config/initializers/simple_form.rb
```diff
@@ -52,21 +28,21 @@
...
   # Use :to_sentence to list all errors for each field.
-  config.error_method = :to_sentence
+  # config.error_method = :first
 
   # Default tag used for error notification helper.
   config.error_notification_tag = :div
 
   # CSS class to add for error notification helper.
-  config.error_notification_class = 'alert alert-error'
+  config.error_notification_class = 'error_notification'
```
asullivan210: do we need to still apply the `.alert-danger` styles we were getting with `alert-error`?

config/initializers/simple_form.rb
```diff
@@ -92,13 +68,14 @@
   # config.item_wrapper_class = nil
 
   # How the label text should be generated altogether with the required text.
-  # config.label_text = lambda { |label, required| "#{required} #{label}" }
+  config.label_text = lambda { |label, required, explicit_label| "#{} #{label}" }
```
asullivan210: so this matches the default just with no indication of whether it is required? probably out of scope but curious if we should add required text to i18n and just use it if it is encouraged
https://github.com/plataformatec/simple_form/blob/master/lib/simple_form.rb#L87

app/views/admin/bulk_assignments/new.html.erb
```diff
@@ -25,16 +23,13 @@
         <%= f.input :force_identifier_creation, as: :boolean, label: 'YES', wrapper_html: { class: "has-error" } %>
       </div>
     <% end %>
-
-    <div class="InfoCard-row">
+    <div class="IdentifierTypeSelection sc-m-v--s">
       <%= f.input :identifier_type_id, label: 'Identified By', collection: identifier_types, label_method: :name, value_method: :id, include_blank: true %>
     </div>
-    <div class="InfoCard-row VariantOptions">
+    <div class="VariantOptions">
```
asullivan210: are there styles for this class?

config/initializers/simple_form.rb
agirlnamedsophia: Those were for the styles that bootstrap gave us, so removing bootstrap means no more `.alert-danger`

app/assets/stylesheets/components/_inputs.scss
agirlnamedsophia: ah, I think these were legacy hacks to handle some weirdness with bootstrap. will remove!

app/views/admin/bulk_assignments/new.html.erb
agirlnamedsophia: No, this was for testing purposes only.

agirlnamedsophia:
nanda?

bump @asullivan210

app/views/admin/bulk_assignments/new.html.erb
asullivan210: ok, we have in the past used `fs-` as a prefix for feature spec for these, but maybe something more explicit would be helpful, just a suggestion I think it's cool to leave alone too

config/initializers/simple_form.rb
asullivan210: if we are going to keep using it, might make sense to style it ourselves but since we haven't so far might be cool to punt on that

asullivan210:
domainLGTM

app/views/admin/bulk_assignments/new.html.erb
agirlnamedsophia: Oh, I like that.

agirlnamedsophia:
nanda?

app/assets/stylesheets/typography/_icon_font.scss
```diff
@@ -0,0 +1,29 @@
...
+  font-style: normal;
+  font-weight: normal;
+  font-variant: normal;
+  text-transform: none;
+  line-height: 1;
+
+  /* Better Font Rendering =========== */
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
```
samandmoore: Did this come straight from some library docs?

I ask because we should put this in vendor/assets instead of in our app/assets. The latter is for source code.

vendor/gems/style-closet/app/assets/javascripts/style_closet.js
```diff
@@ -1 +1 @@
-//= require ./style_closet/
+//= require_tree .
```
samandmoore: Is this what's in style closet? Why did this change?

app/assets/fonts/icomoon.svg
```diff
@@ -0,0 +1,29 @@
...
+</metadata>
+<defs>
+<font id="icomoon" horiz-adv-x="1024">
+<font-face units-per-em="1024" ascent="960" descent="-64" />
+<missing-glyph horiz-adv-x="1024" />
+<glyph unicode="&#x20;" horiz-adv-x="512" d="" />
+<glyph unicode="&#xe900;" glyph-name="chevron-thin-down" data-tags="chevron-thin-down" d="M891.802 660.019c13.926 13.722 36.301 13.722 50.125 0s13.875-35.891 0-49.613l-404.89-400.896c-13.824-13.722-36.198-13.722-50.125 0l-404.89 400.896c-13.824 13.722-13.824 35.891 0 49.613 13.875 13.722 36.301 13.722 50.125 0l379.853-365.619 379.802 365.619z" />
+<glyph unicode="&#xe901;" glyph-name="close" data-tags="x" horiz-adv-x="640" d="M640 640l-128 128-192-192-192 192-128-128 192-192-192-192 128-128 192 192 192-192 128 128-192 192 192 192z" />
+<glyph unicode="&#xea10;" glyph-name="checkmark" data-tags="checkmark, tick, correct, accept, ok" d="M864 832l-480-480-224 224-160-160 384-384 640 640z" />
+</font></defs></svg>
```
samandmoore: Same thing here about putting this in vendor. I'm pretty sure this will work if it's in vendor/assets too.

samandmoore:
<<domain tafn

Almost thereeeeee

vendor/gems/style-closet/app/assets/javascripts/style_closet.js
agirlnamedsophia: It changed because I did it wrong to begin with (the input PR) and I didn't catch it until this PR because the new inputs weren't able to be used until bootstrap was properly deprecated.

agirlnamedsophia:
nanda?

samandmoore:
Domainlgtm

Issue #38, opened 2017-03-30
Heroku deployment does not work as specified


thestumonkey:
I clicked the deploy to heroku button and when I start the dynos, this is in the logs:
```
2017-03-30T10:05:04.141611+00:00 heroku[web.1]: Process exited with status 1
2017-03-30T10:05:59.921567+00:00 heroku[web.1]: State changed from crashed to starting
2017-03-30T10:06:04.032458+00:00 heroku[web.1]: Starting process with command `bundle exec puma -C config/puma.rb`
2017-03-30T10:06:06.829457+00:00 heroku[web.1]: Process exited with status 1
2017-03-30T10:06:06.839395+00:00 heroku[web.1]: State changed from starting to crashed
2017-03-30T10:06:06.550449+00:00 app[web.1]: fatal: Not a git repository (or any parent up to mount point /app)
2017-03-30T10:06:06.550462+00:00 app[web.1]: Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
2017-03-30T10:06:06.553025+00:00 app[web.1]: fatal: Not a git repository (or any parent up to mount point /app)
2017-03-30T10:06:06.553026+00:00 app[web.1]: Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
2017-03-30T10:06:06.683988+00:00 app[web.1]: bundler: failed to load command: puma (/app/vendor/bundle/ruby/2.2.0/bin/puma)
2017-03-30T10:06:06.684016+00:00 app[web.1]: KeyError: key not found: "PUMA_THREAD_COUNT"
2017-03-30T10:06:06.684017+00:00 app[web.1]:   config/puma.rb:11:in `fetch'
2017-03-30T10:06:06.684018+00:00 app[web.1]:   config/puma.rb:11:in `_load_from'
```

jmileham:
Thanks for reporting. We'll get it patched up ASAP. I wonder if there's any way that we could regression protect this feature with an automated test... obviously we don't stand up a fresh heroku instance every day.

thestumonkey:
That is literally the fastest I've ever seen one of my reported issues resolved.  Do I need to delete my heroku app and press the button again?

jmileham:
That would be an easy way to go if you're not invested in that instance, or you can push up a fresh copy of `master` to your running app via git and it should Just Work™.  Let me know if you have any trouble. Marking as closed for now!

Issue #12, opened 2016-11-02
Upgrade to latest version of Puma and revisit configuration


samandmoore:
Currently, we don't have a Procfile for deployments to heroku, so heroku will use the default configuration for running the web server.

That's probably good though because the current version of our `config/puma.rb` probably isn't really heroku compatible. We should consider dropping in a different version of that file or removing it altogether if it's not really needed.

jmileham:
`config/puma.rb` is now compatible with Heroku.

PR #39, opened 2017-03-30
Fix Heroku button

jmileham:
### Summary

The Heroku button regressed because our puma.conf (which used to be ignored but seems to be used by Heroku now) is not aligned with a Heroku 12-factor app - it uses unconventional thread/worker variable names and binds to a unix socket instead of a TCP port. I made switching to TCP conditional on the presence of a PORT env var, which Heroku provides, and added fall-thru to the heroku conventional thread scaling variable names. In the process I swapped the ordering of the preload to be more conventional with Heroku's recommended config. I'm not sure whether this will cause trouble for our production setup at Betterment. We'll want to do a stage deploy after merge and possibly a follow-on patch if it doesn't deploy correctly in our env anymore.

### Other Information

Thanks @thestumonkey for reporting #38

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -1,19 +1,20 @@
-rails_env = ENV['RAILS_ENV'] || ENV['RACK_ENV'] || 'development'
-environment rails_env
+workers Integer(ENV['PUMA_WORKER_COUNT'] || ENV['WEB_CONCURRENCY'] || 2)
+thread_count = Integer(ENV['PUMA_THREAD_COUNT'] || ENV['RAILS_MAX_THREADS'] || 5)
+threads thread_count, thread_count
 
-if rails_env == 'development'
-  thread_count = Integer(ENV.fetch('PUMA_THREAD_COUNT', '5'))
-  threads thread_count, thread_count
+preload_app!
+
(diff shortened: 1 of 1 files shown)


jmileham:
Testing-wise, I tested by pushing this branch to a fresh heroku install with no special config variables set and verified that it works.  I think the only way to fully regression protect against this kind of thing would be a rather anti-social smoke test that literally clicks the heroku button and stands up a temporary app. No doubt a violation of terms.

We could look at a smaller step of [validating the app.json file itself](https://github.com/app-json/app.json#validating-a-manifest), but that's not what bit us here - it was depending on an ENV var not set by the app.json at runtime.

But really, I think regressions like this should be rare.

samandmoore:
this probably addresses part of this issue too https://github.com/Betterment/test_track/issues/12

jmileham:
I also double checked by pushing the heroku button on my branch, which now works.

samandmoore:
> But really, I think regressions like this should be rare.

👍  yup, i think doing nothing re: testing is the right move for now.

samandmoore:
<<domainlgtm

PR #41, opened 2017-03-30
SplitDetail model and additional fields

agirlnamedsophia:
### Summary
* this wraps split details in another model so we can validate these fields only when they're being edited in this form
* this PR also includes two new fields, location and platform


app/models/split.rb
```diff
@@ -19,6 +19,8 @@ class Split < ActiveRecord::Base
 
   scope :active, -> { where(finished_at: nil) }
 
+  enum platform: [:mobile, :desktop]
+
   def has_details?
     %w(hypothesis assignment_criteria description owner).any? { |attr| public_send(attr).present? }
```
noveng05: do we need to add `location` and `platform` here too?

app/models/split_detail.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def location
+    @location ||= split.location
+  end
+
+  private
+
+  def update_split!
+    split.update!(
```
noveng05: Just for my understanding, these attributes you're setting are the `split_details`, and there really shouldn't be any case where the existing `split` is invalid right?

app/models/split_detail.rb
```diff
@@ -0,0 +1,58 @@
+class SplitDetail
+  include ActiveModel::Model
+
+  attr_accessor :split, :hypothesis, :assignment_criteria, :description, :owner, :location, :platform
```
noveng05: for all the attributes other than `:split`, can we use `delegate_attributes` for them?
If we do that, I believe we can remove all the getter methods in here and omit all the params in `split.update!`. 
We _may_ just need to make sure we grab the `split` param in the initialize before the other attributes tho:
```ruby
def initialize(params)
  raise ...
  self.split = params.delete(:split)
  super
end
```
I'm not 100% confident if this would be the right way of using `delegate_attributes`, so would like @samandmoore or anyone else to confirm

agirlnamedsophia:
nanda?

agirlnamedsophia:
bump @noveng05

app/models/concerns/delegate_attribute.rb
```diff
@@ -0,0 +1,54 @@
+module DelegateAttribute
```
noveng05: ah, okay, forgot that test_track is open source. Question for @Betterment/test_track_core. In this case, do we prefer to just use `delegate` rather than `delegate_attribute`
```ruby
delegate :hypothesis, :hypothesis=, :assignment_criteria, :assignment_criteria=, :description, :description=, :owner, :owner=, :location, :location=, :platform, :platform=, to: :split
```

agirlnamedsophia: `delegate` doesn't work like `delegate_attribute` does, and talked to @jmileham and we are fine including this module here for now.

noveng05:
<< platformlgtm
I would say my review was more of a platform review. I'd feel more comfortable if someone who's more close to the test track domain can take one more quick look at it! @Betterment/test_track_core

dschaub:
<< domain lgtm platform lgtm

PR #43, opened 2017-04-05
API for split details

agirlnamedsophia:
This endpoint exposes split details.


jmileham:
<< domain

app/controllers/api/v1/split_details_controller.rb
```diff
@@ -0,0 +1,17 @@
+class Api::V1::SplitDetailsController < UnauthenticatedApiController
+  include CorsSupport
+
+  def show
+    if split.present?
+      @split_detail = SplitDetail.new(split: split)
+    else
+      render json: { error: "Split not found" }, status: :unprocessable_entity
+    end
```
dschaub: any reason we're not 404ing here?

app/views/api/v1/split_details/show.json.jbuilder
```diff
@@ -0,0 +1,7 @@
+json.split_name @split_detail.split.name
+json.hypothesis @split_detail.hypothesis
+json.assignment_criteria @split_detail.assignment_criteria
+json.description @split_detail.description
+json.owner @split_detail.owner
+json.location @split_detail.location
+json.platform @split_detail.platform
```
dschaub: there's a shorthand you can use for pulling values and assigning them to keys of the same name:

```
json.(@split_detail, :hypothesis, :assignment_criteria, ...)
```

jmileham:
@jmileham >> domain

app/controllers/api/v1/split_details_controller.rb
```diff
@@ -0,0 +1,5 @@
+class Api::V1::SplitDetailsController < AuthenticatedApiController
+  def show
+    @split_detail = SplitDetail.new(split: Split.find_by!(name: params[:id]))
```
samandmoore: So we have two-ish places that are doing this now. Should we go ahead and make a reader on the `Split` model for `detail`? then this would become `split.find_by!(...).detail` and the controller for editing split details would be able to use the same pattern. Does that makes sense to you?

app/models/split_detail.rb
```diff
@@ -3,7 +3,7 @@ class SplitDetail
   include DelegateAttribute
 
   attr_accessor :split
-  delegate_attribute :hypothesis, :assignment_criteria, :description, :owner, :location, :platform, to: :split
+  delegate_attribute :name, :hypothesis, :assignment_criteria, :description, :owner, :location, :platform, to: :split
```
samandmoore: should we be able to edit the `name` on this? i don't think we want to make this a `delegate_attribute` but rather just a `delegate`. The difference being that `delegate_attribute` will also delegate the setter allowing changes to be made.

config/routes.rb
```diff
@@ -28,6 +28,7 @@
       resource :assignment_override, only: :create
 
       # Server-side authenticated endpoints
+      resources :split_details, only: :show, controller: 'split_details'
```
samandmoore: I don't think we need the `controller:` argument here. Does it not work without it?

samandmoore:
<<domain tafn because dan is taking too long :trollface:

agirlnamedsophia:
bump @samandmoore

app/models/split.rb
```diff
@@ -1,4 +1,5 @@
 class Split < ActiveRecord::Base
+  attr_reader :split_detail
```
dschaub: did you mean to leave this here?

app/models/split.rb
```diff
@@ -22,6 +23,10 @@ class Split < ActiveRecord::Base
 
   enum platform: [:mobile, :desktop]
 
+  def detail
+    SplitDetail.new(split: self)
```
dschaub: this can be memoized

agirlnamedsophia:
nanda?

dschaub:
@samandmoore >> domain
<< domain lgtm

dschaub:
nanda?

PR #44, opened 2017-04-05
API for exposing a visitor's assignment details

dschaub:
### Summary

In order to provide other applications with human-readable information about splits, we want to expose data for a single visitor's assigned variants.


app/controllers/api/v1/variant_details_controller.rb
```diff
@@ -0,0 +1,5 @@
+class Api::V1::VariantDetailsController < AuthenticatedApiController
+  def index
+    @variant_details = Visitor.find(params[:visitor_id]).assignments.includes(:variant_details).map(&:variant_detail)
```
dschaub: so, this was the result of a bit of back and forth about how best to optimize the queries for getting `VariantDetails`s for an assignment. an assignment has a split ID and a variant, which means it should be able to join directly to a `VariantDetail`, but I couldn't get ActiveRecord to do a compound join.

@jmileham any ideas?

dschaub:
ack, hold off on this, left some things out that I was waiting for until sophia's other data was available.

dschaub:
ok, bump, added that additional information and renamed the controller, since it's describing a visitor's assignment and not just a variant.

agirlnamedsophia:
<< domain

app/models/variant_detail.rb
```diff
@@ -8,6 +8,10 @@ def display_name
     super || variant
   end
 
+  def description
+    super || "#{split.name} = #{variant}"
+  end
```
dschaub: toying with the idea of a default description, but not sure what it should say. seems better than saying nothing, since everything will be blank otherwise to start, and at least for our use case, we're not expecting to go back and add metadata to _every_ previous active test.

dschaub: meh, on second thought, bad idea.

spec/requests/api/v1/assignment_details_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'rails_helper'
+
+describe Api::V1::AssignmentDetailsController, type: :request do
+  let(:split) { FactoryGirl.create(:split, name: 'excellent_feature', registry: { enabled: 50, disabled: 50 }, location: 'Home page') }
+
```
agirlnamedsophia: this needs some basic auth, see [here](https://github.com/Betterment/test_track/pull/43/files#diff-fb80641b93ea2ec5501b64f18b9d2d02R9) for an example

dschaub:
bump

dschaub:
bump

agirlnamedsophia:
domain LGTM

app/controllers/api/v1/visitor_details_controller.rb
```diff
@@ -0,0 +1,11 @@
+class Api::V1::VisitorDetailsController < AuthenticatedApiController
+  def show
+    @assignments = VisitorLookup.new(index_params).visitor.assignments.includes(:variant_details)
+  end
+
+  private
+
+  def index_params
```
samandmoore: show params?

samandmoore: or just lookup_params

app/controllers/api/v1/visitor_details_controller.rb
```diff
@@ -0,0 +1,11 @@
+class Api::V1::VisitorDetailsController < AuthenticatedApiController
+  def show
+    @assignments = VisitorLookup.new(index_params).visitor.assignments.includes(:variant_details)
```
samandmoore: a bit of a law-of-demeter violation on this one, but it's not killing anyone.

i think that if this "visitor detail" concept grows at all then we will wanna spawn a little model for this that can internalize these nuanced access patterns bits -- i.e. the surprising includes call and the defaulting that we do inside of Assignment for a non-existent variant detail record.

app/views/api/v1/visitor_details/_assignment_detail.json.jbuilder
```diff
@@ -0,0 +1,5 @@
+json.split_location assignment.split.location
+json.split_name assignment.split.name
```
samandmoore: is this an n+1 for the split? should we be doing `includes(:split)` in the controller?

app/controllers/api/v1/visitor_details_controller.rb
dschaub: whoops. so many renames

app/views/api/v1/visitor_details/_assignment_detail.json.jbuilder
dschaub: fixed

PR #46, opened 2017-04-12
Properly includes variant details by way of the split presenter

agirlnamedsophia:
This fixes a bug where we were exposing an empty array of variant details that had not been updated with new display names and descriptions, and uses the `SplitPresenter` class to set the `variant_details` on the `SplitDetail` model. 

### DIFF
--- a/app/models/split_detail.rb
+++ b/app/models/split_detail.rb
@@ -4,7 +4,7 @@ class SplitDetail
   include DelegateAttribute
 
   attr_accessor :split
-  delegate :name, :variant_details, to: :split
+  delegate :name, :variants, to: :split
   delegate_attribute :hypothesis, :assignment_criteria, :description, :owner, :location, :platform, to: :split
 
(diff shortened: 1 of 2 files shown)


dschaub:
<< domain lgtm

app/models/split_detail.rb
```diff
@@ -16,6 +16,12 @@ def initialize(params)
     super
   end
 
+  def variant_details
+    @variant_details ||= variants.map do |variant|
+      VariantDetail.find_or_initialize_by(split: split, variant: variant)
+    end
+  end
```
samandmoore: i think this code is also nearly identical to the code we added to the SplitPresenter:

```ruby
class SplitPresenter < SimpleDelegator
  def variant_details
    variants.map do |variant|
      detail = VariantDetail.find_or_initialize_by(split: __getobj__, variant: variant)
      VariantPresenter.new(detail)
    end
  end
end
```

should we redefine the presenter in terms of this?

PR #47, opened 2017-04-13
Refactor Variant Detail presentation

agirlnamedsophia:
* Split presenter goes away and VariantDetail handles view related details
* This resolves a problem with the `test_track_rails_client` where we were leaning on updated records to display variant_detail `display_name`s, though on the server we were lazily initializing them. The client would only display `variant_details` that had been updated, so unless the data was updated manually the rows would be empty.


dschaub:
i thought there was a split presenter, or maybe variant presenter, spec that would need to be deleted?

agirlnamedsophia:
@dschaub  you were right! i deleted them

dschaub:
we probably want those specs in variant_detail_spec now

agirlnamedsophia:
bump @dschaub

dschaub:
<< domain lgtm

PR #48, opened 2017-04-20
Small cleanups

agirlnamedsophia:
This cleans up some domain specific language we had in the view. This also adjusts the split details edit page with some increased margin to make that view consistent with the others. And this removes a duplicate field on the footer in one view.


dschaub:
<< domain lgtm

PR #49, opened 2017-05-03
Add screenshot uploads to variant details

dschaub:
### Summary

A picture's worth a thousand words, so while it's nice we can describe variants, it would be extra nice to be able to show what they look like. Here, I'm adding paperclip and a `screenshot` attribute to variant details to accomplish just that.

Note that screenshots are disabled by default in all environments except for `test`. They're enabled explicitly by providing an `ATTACHMENT_STORAGE` environment variable. The only options so far would be `local` or `s3`, which requires that AWS credentials and an S3 bucket be configured in the environment.


lib/test_track/attachment_settings.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    def max_size
+      ENV['ATTACHMENT_MAX_SIZE'] || 512.kilobytes
+    end
+
+    private
+
+    def s3_enabled?
+      storage_strategy == 's3' && s3_settings_valid?
```
jmileham: Should this do something squawkier if storage strategy is s3 and the s3 settings aren't valid? Maybe it starts up ok but drops an error in the log? If we did that, we'd probably want it (or the caller) to be memoized so it doesn't happen more than once per app initialization.

spec/factories/variant_details.rb
```diff
@@ -4,5 +4,6 @@
     variant { split.variants.first }
     display_name 'Great variant'
     description 'Really, everyone loves it'
+    screenshot_file_name nil
```
jmileham: is this required or would it be the default anyway?

spec/requests/api/v1/split_details_spec.rb
```diff
@@ -21,7 +21,8 @@
         split: split_with_details,
         variant: 'disabled',
         display_name: 'fantastic_split_with_information is off',
-        description: 'This feature makes nothing happen.'
+        description: 'This feature makes nothing happen.',
+        screenshot_file_name: 'screenshot.jpg'
```
jmileham: This seems like an awkward way to stub in an attachment. I think you should probably be using the screenshot setter with an actual fixture file if you want to test that the paperclip integration works as expected?

spec/factories/variant_details.rb
dschaub: nope, it can go away

lib/test_track/attachment_settings.rb
dschaub: yeah i like the idea of logging an error

jmileham:
<< domain tafn

dschaub:
bump

dschaub:
we probably want a mention of this somewhere in the readme, right?

db/schema.rb
```diff
@@ -158,12 +158,16 @@
...
-    t.uuid     "split_id",     null: false
-    t.string   "variant",      null: false
-    t.string   "display_name", null: false
-    t.text     "description",  null: false
-    t.datetime "created_at",   null: false
-    t.datetime "updated_at",   null: false
+    t.uuid     "split_id",                null: false
+    t.string   "variant",                 null: false
+    t.string   "display_name"
+    t.text     "description"
```
samandmoore: were these two things supposed to change to nullable?

dschaub: yikes, how did that happen...

spec/features/admin_variant_details_edit_spec.rb
```diff
@@ -21,11 +23,15 @@
...
+    variant_page.form.screenshot.set variant_screenshot
     variant_page.form.submit_button.click
 
     expect(split_page).to be_displayed
     expect(split_page).to have_content 'Details for enabled have been saved'
     expect(split_page.variants.first.name).to have_content "Variant name"
     expect(split_page.variants.first.description).to have_content "Super great variant"
+
+    split_page.variants.first.edit_link.click
+    expect(variant_page.form.current_screenshot).to have_content 'ttlogo.png'
```
jmileham: So if you are using real attachments in tests, we'll want to make sure that we clean out the test environment's uploads directory between test runs. Does that happen already?

dschaub: no it wasn't, I added some extra config to test.rb to put uploads in tmp and added a cleanup hook in rails_helper

README.md
```diff
@@ -117,6 +117,19 @@ In order to access the admin features of the TestTrack server, you must create a
...
 ```
 
+### Enabling attachments
+Variants can be associated with metadata to describe their effects in human-readable terms. To enable variant screenshots, you'll need some additional configuration in environment variables.
+
+- To store uploads on the local file system, set `ATTACHMENT_STORAGE` to `local`, and:
+  - `LOCAL_UPLOAD_PATH` (optional, defaults to `:rails_root/public/system/:class/:attachment/:id_partition/:style/:filename`)
+- To store uploads in S3, set `ATTACHMENT_STORAGE` to `s3`, and:
+  - `S3_ACCESS_KEY_ID` (required)
+  - `S3_SECRET_ACCESS_KEY` (required)
```
jmileham: Hmmm, I'm just realizing that we should maybe be using the AWS default Env var names for these? If you use their defaults, the lib doesn't need to be passed them at all. Especially in a single-purpose app like TestTrack it doesn't make much sense to have multiple different IAM identities associated with the app, but I believe in general there's no need for an app to have more than one identity either. Can we make these the expected AWS client default var names?

jmileham:
tafn, sorry I keep noticing tinier and tinier details. I suspect we can remove the whole s3_credentials subtree of the paperclip config if we get the env var names matching the AWS lib's expectations. But perhaps Paperclip doesn't make that easy, in which case it still makes sense to at least use the standard variable names and plumb them through manually, just for convention reasons.

README.md
dschaub: it looks like paperclip requires that credentials be passed and won't lean on aws-sdk figuring it out for itself (not that I can tell, anyway). but yeah, I can change these to be the same names as aws-sdk would expect anyway, so `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`

dschaub: (and AWS_REGION)

jmileham:
domain lgtm!

PR #55, opened 2017-10-25
add jruby to travis

### LINKED ISSUE #53: test against latest jruby
update .travis.yml

gregmolnar:
closes #53

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: ruby
 rvm:
 - 2.3.1
+- jruby-9.1.9.0
 addons:
   postgresql: '9.4'
 before_install:
--- a/Gemfile
+++ b/Gemfile
@@ -74,7 +74,7 @@ end
 
 group :development do
   # Access an IRB console on exception pages or by using <%= console %> in views
-  gem 'web-console', '~> 2.0'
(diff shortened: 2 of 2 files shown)


smudge:
Hmm -- Travis failed to install jruby. Try specifying a version, like `jruby-9.1.9.0`.

smudge:
Oh I see -- yeah the pg database adapter is MRI-only too. You'll need to use `activerecord-jdbcpostgresql-adapter` on jruby.

samandmoore:
Perhaps, it's not worth it for us to get this app deployable via JRuby. I added the issue because it seemed nice to enable folks to deploy this as a JRuby app if they so desired, but I'm not sure that we want to commit to maintaining JRuby support for this application. 

Given that, and the small, but not insignificant, trouble of getting JRuby support, I'm going to close this PR.

Thanks a lot to @gregmolnar for diving into this issue. I'm sorry for the unsatisfying result. I hope that you're still interested in getting involved in the project, if so, check out some more of issues on GitHub.

gregmolnar:
No worries. I agree, there might be other compatibility issues as well with jruby so it probably doesn't worth the effort.

PR #57, opened 2017-10-31
Less config needed when you use the rails gem

jmileham:
### Summary

In my haste to get all the upgrades working, I included the bare `factory_bot` gem which required more config than `factory_bot_rails` does. So I fixed it.

/domain @aburgel 
/no-platform

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -66,7 +66,7 @@ group :development, :test do
   gem 'selenium-webdriver'
   gem 'database_cleaner'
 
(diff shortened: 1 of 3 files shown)


aburgel:
<< domainlgtm

jmileham:
The failing push build is because I accidentally pushed my branch to the wrong repo the first time and it couldn't be fetched. The PR build proves that we're all good here. Merging.

PR #58, opened 2017-12-11
Rails 5.0

### LINKED ISSUE #21: Upgrade to Rails 5
The title says it all.

smudge:
### Summary

Making the changes required to run/boot the app on Rails 5. I opted to do the upgrade all at once rather than adding a 2nd `.rails5` gemfile, but if y'all would prefer to be able to conditionally boot against multiple Rails versions, I can revert the main gem/lockfiles and spin up a 2nd set.

This fixes #21.

### Other Information

I also addressed a handful of deprecation warnings (see individual commit messages). Some seem to be coming from included gems, so rather than upgrading more gems all at once I left those deprecation warnings in place.

/platform @jmileham @samandmoore @aburgel @chrislopresto

screen shot 2017-12-11 at 5 55 08 pm


app/controllers/admin/bulk_assignments_controller.rb
```diff
@@ -14,7 +14,7 @@ def create
 
   def enqueue_assignments
     if @bulk_assignment_creation.valid?
-      BulkAssignmentJob.perform_later(create_params)
+      BulkAssignmentJob.perform_later(create_params.to_h)
```
smudge: `ActionController::Parameters` no longer inherits from `Hash`, so ActiveJob doesn't know how to serialize it.

app/controllers/application_controller.rb
```diff
@@ -1,4 +1,6 @@
 class ApplicationController < ActionController::Base
+  protect_from_forgery prepend: true
```
smudge: This one was interesting. We call `skip_before_action :verify_authenticity_token` in `UnauthenticatedApiController`, but nothing ever actually defined a `verify_authenticity_token` before_action. In Rails 4, skipping an undefined filter method would silently do nothing, but on Rails 5 it now raises an exception.

So I could either remove the `skip_before_action :verify_authenticity_token` line (all tests pass if I do that), or add this line which actually does add a `verify_authenticity_token` method for us to skip. 😅

app/models/split_creation.rb
```diff
@@ -21,7 +21,7 @@ def self.create(params)
   end
 
   def weighting_registry=(registry)
-    @weighting_registry = registry.each_with_object({}) { |(variant_name, weight), r| r[variant_name.to_s] = weight }
+    @weighting_registry = registry.to_h.stringify_keys
```
smudge: Is this doing anything more than just stringifying keys? I already added the `to_h` because ActionController::Parameters is getting slimmed down (won't inherit from Hash).

app/controllers/application_controller.rb
jmileham: seems good.

app/models/split_creation.rb
jmileham: yeah, seems correct.

config/application.rb
```diff
@@ -29,13 +29,12 @@ class Application < Rails::Application
...
-    # Do not swallow errors in after_commit/after_rollback callbacks.
-    config.active_record.raise_in_transactional_callbacks = true
-
     config.log_tags = [:host, :uuid]
 
     config.cache_store = :memory_store
 
     config.active_job.queue_adapter = :delayed_job
+
+    config.action_controller.raise_on_unfiltered_parameters = true
```
smudge: This will be the default in Rails 5.1 -- basically, in the old world, there are a bunch of Hash methods you could call on a params hash (`each_with_index`, `merge`, etc) that would implicitly ignore strong parameters. In the new world, ActionController::Parameters does not inherit from Hash, and if you cast it to a hash in any way without strong params, it'll blow up.

I also had to upgrade the airbrake gem because it was doing exactly that internally when reporting errors.

samandmoore: Word of caution with upgrading airbrake. They changed their method definitions in a few places at some point.

samandmoore: Specifically I think they changed the attribute name for the API key and possibly the notify_or_ignore method.

jmileham:
@smudge - thoughts on the hard airbrake upgrade? That's a lot of major versions.

samandmoore:
I think we should do it, but we just have to make sure that the ruby gem surface area that we're using is still there.

smudge:
@samandmoore lolwhoops -- so I just pushed an alternate approach that keeps us on v4, but it requires a monkey patch. There's a later release of 4.X that [attempts to fix the issue](https://github.com/airbrake/airbrake/blob/v4.3.2/lib/airbrake/rails/controller_methods.rb) but it in fact does not:

```ruby
def to_hash(params)
  # Rails <= 4
  return params.to_hash if params.respond_to?(:to_hash)

  # Rails >= 5
  params.to_unsafe_h
end
```

The issue is that `params` still responds to `to_hash`. It just blows up if you haven't filtered with strong params yet.

Looks like there are [some bigger changes from v4 to v5](https://github.com/airbrake/airbrake/blob/a1aec06b5d515aba03ee618b37595dea22f8d7bd/docs/Migration_guide_from_v4_to_v5.md). But then after that I don't see many API changes. So if we want to make the leap, we'll have to follow that migration guide.

smudge:
@jmileham @samandmoore moved to a slightly newer version of `airbrake` to make the params workaround simpler. Not seeing any obvious breakages in [the 4.1.0 -> 4.3.2 version diff](https://github.com/airbrake/airbrake/compare/v4.1.0...v4.3.2), but would be good to confirm that we can successfully get error alerts from stage.

jmileham:
Awesome. << domain lgtm platform lgtm

PR #59, opened 2017-12-13
Attempt to un-break admin interface

smudge:
/domain @jmileham 
/platform @jmileham


jmileham:
<< domain lgtm platform lgtm :)

PR #60, opened 2017-12-19
Ditch logentries gem, explicitly log to STDOUT in prod/stage

### LINKED ISSUE #18: Remove Logentries in favor of standard 12factor logging
We should remove the `le` gem in favor of logging to STDOUT.

smudge:
### Summary

I think the Rails 5 stuff broke some logging. This is an attempt to fix that, but we should confirm in stage that this is the correct solution.

Also this resolves #18 - we don't need the gem dependency.

/platform @jmileham @aburgel @samandmoore

### DIFF
--- a/config/environments/demo.rb
+++ b/config/environments/demo.rb
@@ -71,5 +71,6 @@ Rails.application.configure do
   # Set to :debug to see everything in the log.
   config.log_level = :info
 
-  config.logger = ActiveSupport::TaggedLogging.new(Le.new(ENV['LOGENTRIES_TOKEN'])) if ENV['LOGENTRIES_TOKEN'].present?
+  # Use a different logger for distributed setups.
+  config.logger = ActiveSupport::TaggedLogging.new(Logger.new(STDOUT))
 end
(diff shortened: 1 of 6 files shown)


config/environments/demo.rb
```diff
@@ -71,5 +71,6 @@
   # Set to :debug to see everything in the log.
   config.log_level = :info
 
-  config.logger = ActiveSupport::TaggedLogging.new(Le.new(ENV['LOGENTRIES_TOKEN'])) if ENV['LOGENTRIES_TOKEN'].present?
+  # Use a different logger for distributed setups.
+  config.logger = ActiveSupport::TaggedLogging.new(Logger.new(STDOUT))
```
samandmoore: can we throw this in `application.rb` since it's the same errwhere?

samandmoore:
<<domain tafn

aburgel:
<< domainlgtm platformlgtm

config/environments/demo.rb
smudge: Well, it's not the same in `test` and `development`. But I could do the inverse -- is there a `Logger.new` for writing to `logs/X.log`?

aburgel: except for `development` and `test`?

samandmoore: oh, right. derp. nvm, let's roll with this.

samandmoore:
domainlgtm

PR #61, opened 2017-12-19
Changes to schema file after Rails 4->5 upgrade

smudge:
### Summary

I never ran `rake db:migrate` when testing on Rails 5, so I didn't notice that this file would change.

So here are the changes!

/domain @Betterment/test_track_core @jmileham @samandmoore 
/no-platform


jmileham:
love that index inlining!

<< domain lgtm!

PR #62, opened 2018-01-24
Add semantic logger with json output

aburgel:
/no-platform
/cc @smudge 

The [rails_semantic_logger](https://github.com/rocketjob/rails_semantic_logger) gem allows you to format log messages using json (among other things), which is much easier to parse. This PR provides an option to use the gem via an environment variable.


jmileham:
<< domain lgtm!

PR #63, opened 2018-01-25
Fix semantic logger with puma

aburgel:
/no-platform

Forgot that we need to add a bit of config for Semantic Logger to work with Puma: https://rocketjob.github.io/semantic_logger/forking.html#puma

The `if defined?` conditional ensures that we don't depend on the library if its not enabled.


aburgel:
nanda?

PR #64, opened 2018-01-31
address race condition in visitor supersession

samandmoore:
### Summary

we occasionally see two supersessions occur concurrently for the same
visitor. one of those will succeed and the other will fail. this results
in the loss of the supersession history. this is not really a huge deal,
but the annoying thing is that this results in an inactionable error.

the solution in this changeset is to use savepoints and rescuing to
swallow the unique violations.

/platform @jmileham @aburgel 

Does anyone know if i definitely do or do not have to capture the `PG::UniqueViolation`? I believe that Rails will wrap that in `ActiveRecord::RecordNotUnique`, but it's hard to reproduce this exact case locally.

### DIFF
--- a/app/models/visitor_supersession.rb
+++ b/app/models/visitor_supersession.rb
@@ -24,7 +24,7 @@ class VisitorSupersession < ActiveRecord::Base
     end
   rescue ActiveRecord::RecordNotUnique, PG::UniqueViolation # rubocop:disable Lint/HandleExceptions
     # the goal here is to make sure that any splits that the old visitor had
-    # that the new visitor doesn't have are carried over to the new visitor.
+    # (that the new visitor doesn't have) are carried over to the new visitor.
     # so, if there is a conflict here, we can safely ignore it.
   end
 end
--- a/app/models/visitor_supersession.rb
+++ b/app/models/visitor_supersession.rb
@@ -23,7 +23,8 @@ class VisitorSupersession < ActiveRecord::Base
       )
     end
   rescue ActiveRecord::RecordNotUnique, PG::UniqueViolation # rubocop:disable Lint/HandleExceptions
-    # ignore or find
-    # we're not using the result, so probably just ignore
+    # the goal here is to make sure that any splits that the old visitor had
+    # that the new visitor doesn't have are carried over to the new visitor.
+    # so, if there is a conflict here, we can safely ignore it.
   end
 end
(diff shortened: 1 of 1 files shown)


app/models/visitor_supersession.rb
```diff
@@ -6,15 +6,24 @@ class VisitorSupersession < ActiveRecord::Base
 
   private
 
-  def merge_assignments! # rubocop:disable Metrics/AbcSize
+  def merge_assignments!
     target_split_ids = superseding_visitor.assignments.map(&:split_id).to_set
-    superseded_visitor.assignments.order(:id).map do |a|
+    superseded_visitor.assignments.order(:id).each do |a|
```
samandmoore: i changed the map to each because we weren't doing anything with the results and this a private method that is called from a lifecycle hook.

app/models/visitor_supersession.rb
```diff
@@ -6,15 +6,24 @@ class VisitorSupersession < ActiveRecord::Base
...
+        variant: assignment.variant,
+        split_id: assignment.split_id,
         context: 'visitor_supersession',
         visitor_supersession: self
-      ) unless target_split_ids.include?(a.split_id)
+      )
     end
+  rescue ActiveRecord::RecordNotUnique, PG::UniqueViolation # rubocop:disable Lint/HandleExceptions
+    # ignore or find
+    # we're not using the result, so probably just ignore
```
jmileham: I'd say why we can ignore - the goal is to fill gaps, not override anything, or read anything.

jmileham:
<< domain lgtm platform lgtm

jmileham:
If we find perf problems from the extra SQL chattiness we can go deeper.

app/models/visitor_supersession.rb
```diff
@@ -6,15 +6,25 @@ class VisitorSupersession < ActiveRecord::Base
...
+        split_id: assignment.split_id,
         context: 'visitor_supersession',
         visitor_supersession: self
-      ) unless target_split_ids.include?(a.split_id)
+      )
     end
+  rescue ActiveRecord::RecordNotUnique, PG::UniqueViolation # rubocop:disable Lint/HandleExceptions
+    # the goal here is to make sure that any splits that the old visitor had
+    # that the new visitor doesn't have are carried over to the new visitor.
+    # so, if there is a conflict here, we can safely ignore it.
```
samandmoore: @jmileham does this sound good?

samandmoore: actually, just updated once more.

jmileham: sgtm

PR #66, opened 2018-03-28
delete invalid and unused delegation

samandmoore:
### Summary

Was reviewing the code and happened to notice that this delegation is being overridden by real methods _and_ would not work if it wasn't being overridden. 

/no-platform


jmileham:
<< domain lgtm!

PR #67, opened 2018-04-04
Semantic Logger improvements

reustmd:
### Summary

1. Propogate conditional usage of taggedLogger based on semantic logger env var. This didn't cause actual symptoms because it was overwritten later, but I think this clarifies the intent.

2. Use the app name to avoid the current default of 'Semantic Logger'

### DIFF
--- a/config/environments/demo.rb
+++ b/config/environments/demo.rb
@@ -71,6 +71,7 @@ Rails.application.configure do
   # Set to :debug to see everything in the log.
   config.log_level = :info
 
-  # Use a different logger for distributed setups.
-  config.logger = ActiveSupport::TaggedLogging.new(Logger.new(STDOUT))
+  if ENV["SEMANTIC_LOGGER_ENABLED"].blank?
+    config.logger = ActiveSupport::TaggedLogging.new(Logger.new(STDOUT))
+  end
 end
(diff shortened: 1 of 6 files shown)


jmileham:
<< domain
/platform @jmileham
<< platform

config/environments/demo.rb
```diff
@@ -71,6 +71,7 @@
   # Set to :debug to see everything in the log.
   config.log_level = :info
 
-  # Use a different logger for distributed setups.
-  config.logger = ActiveSupport::TaggedLogging.new(Logger.new(STDOUT))
+  if ENV["SEMANTIC_LOGGER_ENABLED"].blank?
```
jmileham: Is this blank treatment as false a consistent pattern in TT? I often prefer to use something a little richer in case somebody drops a `0` or `false` into the env var or something: https://stackoverflow.com/a/44322375

But I wouldn't boil the ocean over it if this is common already in TT env-var handling.

jmileham:
one comment maybe worth looking at. platform lgtm domain lgtm

config/environments/demo.rb
reustmd: i agree, i was only following the convention that was in place. i am happy to make it a smidge more explicit.

i'll give other folks a chance to chime in.

/cc @samandmoore @smudge @aburgel

PR #68, opened 2018-04-11
Upgrade deps

jmileham:
### Summary

Upgraded some vulnerable deps. For our use case, they were not exploitable, but now we should be clean again.

### Other Information

I got on the latest rubocop in the process here. There are some wild cops I have some thoughts/questions on that we might want to consider globally? /cc @klesse413

/no-platform


lib/tasks/seed_app.rake
```diff
@@ -1,8 +1,8 @@
 desc "Create a new seed app including a random auth_secret for local development"
 task :seed_app, [:app_name] do |_, opts|
-  seed_app_filename = Rails.root.join('db/seed_apps.yml')
+  seed_app_filename = Rails.root.join('db', 'seed_apps.yml')
```
jmileham: This is my least favorite cop behavior. especially given this: https://twitter.com/tenderlove/status/842064491936280576

The description of this cop says it's doing it to allow for platforms on which separator isn't `/`, which of course according to the above tweet isn't true of any platform in this universe. So it is basically just making your code more verbose.

The path join methods are a nice convenience (I love using them), and ensuring that no escaping magic is used to fool you when concatenating in user input is maybe good (though I'm not sure it'd be sufficient to prevent even `../..`-style attacks), but forcing you not to use slashes in your literal string paths is just for show.

What do you think?

lib/tasks/vendor_deps.rake
```diff
@@ -2,55 +2,9 @@ namespace :test_track do
   task :vendor_deps do
     FileUtils.module_eval do
       cd "vendor/gems" do
-        rm_r Dir.glob('*')
-        `git clone --depth=1 git@github.com:Betterment/style-closet.git && rm -rf style-closet/.git`
```
jmileham: Froze from upgrading style-closet anymore. Lots changed, and it didn't seem worth keeping up?

samandmoore:
<<domain

lib/tasks/seed_app.rake
samandmoore: i feel like we should just disable this cop.

lib/tasks/vendor_deps.rake
samandmoore: seems right to me, not without making updates to the styles

samandmoore:
maybe just disable that path cop.

domainlgtm!

jmileham:
I still like the half of the rails join cop that says not to use the `File.join` when you can use the `Rails.root.join`, but yeah, it's hardly worth it in its current form. Disabling.

app/models/split.rb
```diff
@@ -92,12 +92,12 @@ def variants_must_be_snake_case
   end
 
   def registry_weights_must_sum_to_100
-    sum = registry && registry.values.sum
+    sum = registry && registry.values.sum # rubocop:disable Style/SafeNavigation
```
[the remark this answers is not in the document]
klesse413: hm so https://github.com/bbatsov/rubocop/commit/a1013297f71349f775323367cc48e3863168c869#diff-57a1b37ed7f1de40acdf4f91a7720f49 was released as part of 0.53. but they didn't make it configurable, they only did it for chains _exceeding_ length 2. we could always open a pr to make it configurable (i.e. an option to IgnoreChainedMethods or something like that)

lib/tasks/seed_app.rake
klesse413: fine with disabling, although its enabled in rubocop-betterment

PR #69, opened 2018-04-12
Don't record assignment events for feature gates

jmileham:
### Summary

Because feature gates are not used for analysis, there's no need to record anything but overrides for feature gates. This will vastly reduce the data that needs to change when deciding a feature gate split. It also will cause reweighting feature gates to affect clients in the field instantly unless they've been overridden.

Also explain the difference between experiments and gates to admins.

### Other Information

This is compatible with all clients currently in the field, though behavior will change a bit. The chrome extension will reflect that no matter how many times you've experienced a gate, you are not assigned to a variant unless you assigned it via the extension or admin interface (or before this upgrade took place).

This should make slow rolling features much easier.

This is currently heuristically tagging splits as feature gates based on an `_enabled` suffix. The goal would be to later make creating feature gates (as distinct from experiments) a first class feature of the API so that you provision them differently, and for example, you only get true/false variants from a gate. But this is a very small step code-wise which should hopefully make a big difference to users of splits at scale that otherwise took a lot of dancing  to slow roll.

/domain @ceslami @klesse413 @betterzega or anybody else who has done big slow roll releases who has opinions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,44 @@ One the values that a given visitor will be assigned for a split, e.g. `true` or
 ### Weighting
 Variants are assigned pseudo-randomly to visitors based on their visitor IDs and the weightings for the variants.  Weightings describe the probability of a visitor being assigned to a given variant in integer percentages.  All the variant weightings for a given split must sum to 100, though variants may have a weighting of 0.
 
+### Experiment
+
+Experiments are the standard flavor of splits in TestTrack. They are
+intended to be used for A/B testing, and the TestTrack server records
+visitors' experienced variants so that those visitors will continue to
+experience the same variant regardless of subsequent changes to the
+weightings of those variants via the admin interface.
+
+Storing the variant a visitor experienced for an experiment also allows
+TestTrack to provide a consistent UX to a customer who experienced a
+new-to-them experiment before logging in on a new device, only to be
+recognized as an existing visitor upon sign-in.  TestTrack will merge
+all variant assignments from the anonymous visitor into the
+authenticated visitor at sign-in as long as the authenticated visitor
+doesn't have conflicting assignments.  In that case, the authenticated
+visitor's previous assignments win.
+
+### Feature Gate
+
+As of TestTrack version 1.2, splits with names ending in the `_enabled`
+suffix will be treated as feature gates. Feature gates differ from
+experiments in that they are not intended to be used for analysis, and
+therefor it is not important that the user remain in the same variant
+throughout the entire split lifecycle. Feature gates are meant to be
+slow-rolled (incrementally increasing the percentage of customers
+experiencing the new feature), released en masse, or instantly rolled
+back.
+
+To facilitate these smooth transitions and rapid toggles, the TestTrack
+server will not record variant assignments when a visitor experiences a
+split. This means that every time a visitor experiences a split, they
+will be deterministically (pseudorandomly) assigned to a variant based
+on their visitor ID and the name of the split. This will provide the
+customer with a stable variant given a constant split weighting, but
+probablistically increase the size of the population experiencing the
+the `true` variant as the split weightings are increased via the admin
+panel, giving an admin full control over the feature's release.
+
 ### IdentifierType
 A name for a customer identifier that is meaningful in your application, typically things that people sign up as, log in as.  They should be expressed in `snake_case` and conventionally are prefixed with the application name that the identifier is for, e.g. `myapp_user_id`, `myapp_lead_id`.
 
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ visitor's previous assignments win.
 As of TestTrack version 1.2, splits with names ending in the `_enabled`
 suffix will be treated as feature gates. Feature gates differ from
 experiments in that they are not intended to be used for analysis, and
-therefor it is not important that the user remain in the same variant
+therefore it is not important that the user remain in the same variant
 throughout the entire split lifecycle. Feature gates are meant to be
 slow-rolled (incrementally increasing the percentage of customers
 experiencing the new feature), released en masse, or instantly rolled
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ split. This means that every time a visitor experiences a split, they
 will be deterministically (pseudorandomly) assigned to a variant based
 on their visitor ID and the name of the split. This will provide the
 customer with a stable variant given a constant split weighting, but
-probablistically increase the size of the population experiencing the
+probablistically increase the percentage of visitors experiencing the
 the `true` variant as the split weightings are increased via the admin
 panel, giving an admin full control over the feature's release.
 
--- a/app/models/deterministic_assignment_creation.rb
+++ b/app/models/deterministic_assignment_creation.rb
@@ -14,13 +14,15 @@ class DeterministicAssignmentCreation
   end
 
   def save!
-    ArbitraryAssignmentCreation.create!(
-      visitor_id: visitor_id,
-      split_name: split_name,
-      variant: variant,
-      mixpanel_result: mixpanel_result,
-      context: context
-    )
+    unless split.feature_gate?
+      ArbitraryAssignmentCreation.create!(
+        visitor_id: visitor_id,
+        split_name: split_name,
+        variant: variant,
+        mixpanel_result: mixpanel_result,
+        context: context
+      )
+    end
   end
 
   def variant
--- a/app/models/split_creation.rb
+++ b/app/models/split_creation.rb
@@ -25,7 +25,11 @@ class SplitCreation
   end
 
   def split
-    @split ||= app.splits.create_with(registry: merged_registry).find_or_initialize_by(name: name)
+    @split ||= app.splits.create_with(registry: merged_registry, feature_gate: feature_gate?).find_or_initialize_by(name: name)
+  end
+
+  def feature_gate?
+    name.end_with?("_enabled")
   end
 
   private
--- a/app/views/admin/split_configs/new.html.erb
+++ b/app/views/admin/split_configs/new.html.erb
@@ -5,6 +5,22 @@
   <h1 class="TakeoverText-title sc-m-v--l">Split: <%= @split_creation.name %></h1>
 </div>
 
+<% if @split_creation.feature_gate? %>
+  <p>
+  This split is a feature gate. Changing weights will immediately change
+  behavior of visitors who do not have an explicit assignment, even if they've
+  already experienced a specific variant of this split. This is usually
+  desirable for slow-rolling features.
+  </p>
+<% else %>
+  <p>
+  This split is an experiment. Changing weights will have no immediate effect
+  on the behavior of visitors who have already experienced a variant this
+  split. Experiments rarely benefit from changing weightings unless you are
+  performing analysis over a date range.
+  </p>
+<% end %>
+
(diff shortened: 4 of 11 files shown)


db/migrate/20180412153251_add_split_feature_gate.rb
```diff
@@ -0,0 +1,6 @@
+class AddSplitFeatureGate < ActiveRecord::Migration[5.0]
+  def change
+    add_column :splits, :feature_gate, :boolean, default: false, null: false
```
jmileham: This is a high throughput table but very small number of rows and there are effectively no long-running transactions on this service. My hope is this doesn't cause issues on deploy. :) /cc @Betterment/data for commentary.

app/models/split_creation.rb
```diff
@@ -25,7 +25,11 @@ def weighting_registry=(registry)
   end
 
   def split
-    @split ||= app.splits.create_with(registry: merged_registry).find_or_initialize_by(name: name)
+    @split ||= app.splits.create_with(registry: merged_registry, feature_gate: feature_gate?).find_or_initialize_by(name: name)
+  end
+
+  def feature_gate?
+    name.end_with?("_enabled")
```
betterzega: Would we want something more explicit on the split to tag the type of split?

betterzega: ah, nvm, I see the flag below

klesse413: hm yeah, long term i would think probably something more explicit (although shorter term this gets the job done). we use this naming convention, but that's not to say others necessarily do. do we recommend it in documentation anywhere?

jmileham: I believe so but it might just be in a blog post or something.

app/views/admin/splits/_details.html.erb
```diff
@@ -33,5 +35,11 @@
         <td>&nbsp;</td>
       </tr>
     </table>
+    <% if @split.feature_gate? %>
+      <p>
+      <a id="gate_population"></a>
+      * Feature gates do not track assignment events and population reflects only visitors assigned to specific variants.
```
asullivan210: 👍 can we make this even more explicit, maybe `only visitors assigned to specific variants via the chrome extension or admin tool` or something to make clear that if 10 Betterment employees have it on it will show population 10 even if the weight is 20% `true` and thousands fo people have seen the `true` experience

app/models/deterministic_assignment_creation.rb
```diff
@@ -14,13 +14,15 @@ def self.create!(params)
...
   def save!
-    ArbitraryAssignmentCreation.create!(
-      visitor_id: visitor_id,
-      split_name: split_name,
-      variant: variant,
-      mixpanel_result: mixpanel_result,
-      context: context
-    )
+    unless split.feature_gate?
+      ArbitraryAssignmentCreation.create!(
```
asullivan210: chatted offline, @jmileham pointed out that if product still wants to know how many people have seen the enabled experience (e.g. we've turned the feature flag on for 20% of people and want to gauge if there are any blocking bugs) we can query redshift's `customer_prod.split_assigned` table because segment will still track assignment events 💯

jmileham:
/domain @joejansen for analytics backend impact (that makes 3 domain reviews, I don't wanna jump the gun on this)

asullivan210:
nanda?

asullivan210:
<<domain

app/models/split_creation.rb
jmileham: Yup, here's the [blog post](https://engineering.betterment.qa/2017/06/22/testtrack-implementation-best-practices.html#naming-split-tests-and-feature-flags) by the man himself @joejansen

app/views/admin/splits/_details.html.erb
jmileham: done.

app/models/split_creation.rb
klesse413: 🆒 no external documentation though right? seems worth highlighting for other users that they'll get this functionality based on naming convention

app/views/admin/split_configs/new.html.erb
```diff
@@ -5,6 +5,22 @@
...
+  <p>
+  This split is a feature gate. Changing weights will immediately change
+  behavior of visitors who do not have an explicit assignment, even if they've
+  already experienced a specific variant of this split. This is usually
+  desirable for slow-rolling features.
+  </p>
+<% else %>
+  <p>
+  This split is an experiment. Changing weights will have no immediate effect
+  on the behavior of visitors who have already experienced a variant this
```
klesse413: a variant _of_ this split?

jmileham: :man_facepalming:

jmileham:
So in the interest of transparency, this will result in duplicate assignment events per visitor for gates experienced more than once. Normally a client retrieves all assignments from TT before displaying any content to the visitor, and if an assignment for a given split exists, it'll just render it. But since the server will have no record of an assignment, every single time a customer experiences a gate, another assignment event will be triggered. This will both hit segment and TT. So it'll increase event volume a bit for a given level of traffic, and if we cared about the event data (which we do not) it would be misleadingly noisey, with duplicate events happening a fair amount.

In future iterations of client libs we can phase out sending assignment events for feature gates and focus on only sending experience events, which we'd want to send each time.

samandmoore:
@jmileham can you spec that out as a GitHub issue in one of the clients like the ruby one? And then link it up in a corresponding issue in the js client? It’d be great to have an record of that.

jmileham:
Another side effect of this strategy is that signing in from a new computer or any other activity that ends up swapping your visitor ID will likely cause you to end up in a different variant for a given feature gate than you would have as your anonymous visitor, even if you've never experienced that feature gate before. Normal visitor supersession logic would've applied the variant from your anonymous visitor to your authenticated visitor if the authenticated visitor had no such assignment, but in this case, you will simply get a deterministic assignment appropriate to your visitor ID each time.

This isn't a big change for most folks. Once you've signed into a device, if you're the only user of the device, your visitor ID will remain constant between sessions so you'll have a consistent experience. And if multiple users use a device, they will have already encountered the same issue as they swap back and forth between their visitor IDs upon login.

README.md
```diff
@@ -154,6 +154,44 @@ One the values that a given visitor will be assigned for a split, e.g. `true` or
...
+
+To facilitate these smooth transitions and rapid toggles, the TestTrack
+server will not record variant assignments when a visitor experiences a
+split. This means that every time a visitor experiences a split, they
+will be deterministically (pseudorandomly) assigned to a variant based
+on their visitor ID and the name of the split. This will provide the
+customer with a stable variant given a constant split weighting, but
+probablistically increase the size of the population experiencing the
+the `true` variant as the split weightings are increased via the admin
+panel, giving an admin full control over the feature's release.
```
jmileham: @klesse413 - I wrote out the definitions and behaviors of experiments and feature gates in the readme. I'm hypothetically calling this version 1.2. It seems big enough to be worth a point release, but I kinda want to save 2.0 for the version that enables a different migration DSL API contract for defining experiments vs gates, and does things like force them to have conventional names and variants.

app/models/split_creation.rb
jmileham: added to readme

README.md
```diff
@@ -154,6 +154,44 @@ One the values that a given visitor will be assigned for a split, e.g. `true` or
...
+authenticated visitor at sign-in as long as the authenticated visitor
+doesn't have conflicting assignments.  In that case, the authenticated
+visitor's previous assignments win.
+
+### Feature Gate
+
+As of TestTrack version 1.2, splits with names ending in the `_enabled`
+suffix will be treated as feature gates. Feature gates differ from
+experiments in that they are not intended to be used for analysis, and
+therefor it is not important that the user remain in the same variant
```
klesse413: therefor 👮

README.md
```diff
@@ -154,6 +154,44 @@ One the values that a given visitor will be assigned for a split, e.g. `true` or
...
+
+To facilitate these smooth transitions and rapid toggles, the TestTrack
+server will not record variant assignments when a visitor experiences a
+split. This means that every time a visitor experiences a split, they
+will be deterministically (pseudorandomly) assigned to a variant based
+on their visitor ID and the name of the split. This will provide the
+customer with a stable variant given a constant split weighting, but
+probablistically increase the size of the population experiencing the
+the `true` variant as the split weightings are increased via the admin
+panel, giving an admin full control over the feature's release.
```
klesse413: won't we only store the assignment if a customer is explicitly assigned via the admin panel? i.e. if we just increase the `true` weighting on incremental rollout, couldn't a customer randomly experience the `true` version and then come back and randomly experience the `false` version? so we _have_ to incrementally release by explicit assignment of customer ids? or am i missing something?

jmileham: The algo works something like this:

truncate_to_integer_and_scale_to_number_between_0_and_99(sha1(visitor_id + split_name))

we then sort the variants alphabetically and then assign them ranges making up the numbers between 0 and 99. You experience the variant that your number buckets to.

So as we ratchet up or down, the same users for a given split will be rolled out to and rolled back from in deterministic, pseudorandom order per split.

That said, I probably shouldn't use the word population in this definition in a way that differs from the population in the admin - none of these visitors will show up in the population in the admin, because they won't have any concrete assignments.

jmileham: (this would work horribly if you had more than two variants in a gate as the middle variant would slide through the population, but what kind of gate has more than two sides? this is why we wanna formalize the API for making gates.)

jmileham:
Alright, I'm being needy.

@klesse413 << domain platform - you can't comment this much without providing a review :)
@joejansen << domain

Would love to land this because it shouldn't be disruptive and will make feature rollouts Much Easier.

app/models/deterministic_assignment_creation.rb
joejansen: excellent. for debugging purposes i think there is value in continuing to fire the `SplitAssigned` event to the analytics client that is configured by the client code.

joejansen:
domainlgtm

app/views/admin/splits/_variants.html.erb
```diff
@@ -1,7 +1,7 @@
 <article class="InfoCard sc-m-v--l">
   <div class="InfoCard-titleContainer">
     <h4>Variant Details</h4>
-    <%= link_to "Edit", new_admin_split_split_config_path(split), class: 'change-weights-link' %>
+    <%= link_to "Change Weights", new_admin_split_split_config_path(split), class: 'change-weights-link' %>
```
klesse413: 🙌 👍 for more descriptive links. ive always been confused by the 5 different "edit" links on the split page - when I want to assign a user I never know which one to click 😂

db/migrate/20180412153251_add_split_feature_gate.rb
klesse413: 👍 not sure what our tt deploy pattern is, maybe we want to deploy to stage first and watch the migration?

klesse413:
domainlgtm platformlgtm 🙌 my team is super excited about this!

db/migrate/20180412153251_add_split_feature_gate.rb
jmileham: Won’t do much for lock contention confidence because of the vastly different load levels but yeah. Will do this and the other off hours unless data says no.

asullivan210:
domainLGTM

PR #71, opened 2018-05-03
Single threaded reassignments

jmileham:
### Summary

Bulk reassignments are not a high priority operation, and in parallel they can saturate available IO for OLTP workloads, causing performance degradation. Since clients have a tight timeout for some requests (as low as 200ms), this can lead to actual problems, though even merely slowing down your end user experience is already not great and a major design goal of TT to avoid. This introduces a dependency on `with_transactional_lock` to ensure that bulk reassignments are never run in parallel even if multiple DJ workers are available to work them, which should reduce IO resource consumption of bulk reassignments to an acceptable level. If it turns out not to be enough, we'll investigate rate limiting batches further.

/no-platform

While I was in there, I got on rubocop-betterment as well. Reviewing the commits in isolation is probably clearer.


app/models/bulk_reassignment.rb
```diff
@@ -7,7 +7,7 @@ class BulkReassignment
 
   def save
     return false unless valid?
-    transaction do
+    ActiveRecord::Base.with_transactional_lock('bulk_reassignment') do
```
samandmoore: we should switch to the modern `ApplicationRecord` thing at some point.

spec/requests/api/v1/split_details_spec.rb
```diff
@@ -55,7 +55,7 @@
           variant: 'disabled',
           display_name: 'fantastic_split_with_information is off',
           description: 'This feature makes nothing happen.',
-          screenshot: File.open(Rails.root.join('spec/support/uploads/ttlogo.png'))
+          screenshot: File.open(Rails.root.join('spec', 'support', 'uploads', 'ttlogo.png'))
```
samandmoore: soooo silly. oh well.

samandmoore:
<<domainlgtm

works for me.

jmileham:
Changed it not to direct ref to activerecord now, delegates to the
underlying model instead.

PR #78, opened 2018-05-18
update database cleaner to 1.7.0

klesse413:
### Summary

Updating database cleaner ahead of upgrading this app to rails 5.1. This is due to a deprecation warning that is accounted for in newer versions of database cleaner [here](https://github.com/DatabaseCleaner/database_cleaner/blob/master/lib/database_cleaner/active_record/truncation.rb#L30).

/domain @betterzega @samandmoore 
/no-platform


samandmoore:
<<domainlgtm

PR #80, opened 2018-06-06
Experience sampling weight

jmileham:
### Summary

Add an EXPERIENCE_SAMPLING_WEIGHT env var and a new v2 `split_registry` endpoint to expose its value to clients.

The idea here is to tell clients to probablistically sample FeatureGateExperienced events at a given sampling weight. Sampling weight is the reciprocal of sample fraction, i.e. a sampling weight of 10 means that one in ten experience events should be reported to the analytics server and hence each sampled event represents ten real-world experiences.

This is a data volume reduction/cost saving measure. The default value is 1, meaning every experience event should be reported. The value 0 is intended to disable all experience events. This relies on the client reading the new split registry endpoint and sampling appropriately.

In order to make this feature work, I had to change the structure of the split_registries endpoint because it didn't have enough flexibility in it. It used to be just a JSON object with keys representing splits. I added a wrapper object and dropped the previous object into a `splits` key. That created space for a `experience_sampling_weight` key.

While I was in there, I actually added more flex to the way split weightings are expressed: there's a similar issue there where the value of a given split was previously just a set of weightings. This means we can't transmit metadata abouta split other than weights in the v1 format. So I'm auditioning an expansion that puts the weightings in a `weights` key and allows us to add things like a `feature_gate` boolean to the payload.

This will add significant verbosity to the page. I'm not sure it's worth it, but it would've allowed me to stop having clients heuristically decide whether a given split is a feature gate or not based on suffix in the last iteration. Thoughts? I can pull that commit out of this PR if we don't want to go there.

/platform @aburgel @samandmoore


app/views/api/v2/split_registries/show.json.jbuilder
```diff
@@ -0,0 +1,10 @@
+json.splits do
+  @split_registry.splits.each do |split|
+    json.set! split.name do
+      json.weights split.registry
+      json.feature_gate split.feature_gate?
+    end
+  end
+  json.merge!({})
```
jmileham: This seems a little ugly but it was the only way I found to force the `splits` key to appear with an empty object as a value. If the loop is empty, otherwise the split key disappears entirely. Is there a more revealing approach?

app/views/api/v2/split_registries/show.json.jbuilder
```diff
@@ -0,0 +1,10 @@
+json.splits do
+  @split_registry.splits.each do |split|
+    json.set! split.name do
+      json.weights split.registry
+      json.feature_gate split.feature_gate?
```
jmileham: Another advantage of this representation would be that we could theoretically set different sample rates for different splits if e.g. different apps were running at different orders of magnitude of scale.

samandmoore:
Sure! I will give it a look tonight.

spec/models/split_registry_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        allow(ENV).to receive(:fetch).and_call_original
+
+        subject.experience_sampling_weight
+        subject.experience_sampling_weight
+
+        expect(ENV).to have_received(:fetch).with('EXPERIENCE_SAMPLING_WEIGHT', any_args).exactly(:once)
+      end
+
+      it "returns 1 with no env var" do
+        with_env EXPERIENCE_SAMPLININGG_WEIGHT: nil do
```
jmileham: Typo

samandmoore:
<<platform

samandmoore:
platformlgtm

jmileham:
Only downside to the compressibility is that presently when we pass the registry through to the JS client in the HTML payload we base64 it first, which will lose all the compressibility. I think the next gen answer here will probably involve GZipping the content before base64ing it or something.

PR #81, opened 2018-06-19
Upgrade to patched sprockets

jmileham:
### Summary

Patch in response to CVE-2018-3760. We do not compile assets in production so the TestTrack app is not vulnerable by default, but best to put this vulnerable version of sprockets in the rearview anyway.

/domain @Betterment/test_track_core @samandmoore @aburgel


samandmoore:
<<domainlgtm

PR #82, opened 2018-07-31
Add app id seed

effron:
### Summary

To go along with this update to the test_track_rails_client https://github.com/Betterment/test_track_rails_client/pull/60, this adds a default "TestTrack" app and an "app_id" identifier type to the seeds file, so that consumer apps all don't have to create their own "app_id" identifier type.

For new apps, the seeds will run as part of `rake db:setup`, and for existing apps, you will have to run `rake db:seed`. It is designed to be idempotent, and accept any already existing apps with the name "TestTrack" or existing identifier types with the name "app_id".

/no-platform


effron:
nanda?

jmileham:
Worth hitting it from a migration as well instead of requiring seeds to be run manually for existing testtrack installs?

db/seeds.rb
```diff
@@ -9,3 +9,11 @@
...
   end
 end
+
+test_track_app = App.find_or_create_by!(name: 'TestTrack') do |app|
+  app.auth_secret = SecureRandom.urlsafe_base64(32)
+end
+
+IdentifierType.find_or_create_by!(name: 'app_id') do |identifier_type|
+  identifier_type.owner_app = test_track_app
+end
```
samandmoore: any reason for this syntax instead of `test_track_app.identifier_types.find_or_create_by!(...)`?

samandmoore:
sure, we can throw a migration in. i suggested we just skip it because we can run the seed command one-off, but no biggie.

do we need to document anything here?

domain tafn

db/seeds.rb
effron: in case someone created an `app_id` identifier type already, then we would get a duplicate key error. since we really only care that any identifier type named "app_id" exists (and don't care about the owner app), this felt like a safe approach

samandmoore: ah yea. good point. sounds good to me.

effron:
I'm not sure if we need to document here, or in test_track_rails_client (or both?). maybe something about when to use app_id, and something about when you should lean on TestTrack.app_ab vs a normal test_track_visitor.test_track_ab

jmileham:
If we skip the migration we gotta add an upgrading section to the ever
growing readme is all.

samandmoore:
Makes sense. I was just thinking we probably need to add a line about
running seeds on initial setup anyway.

effron:
the initial setup recommends running `rake:db:setup` which will run seeds, so i'm nto sure we need to add explicit guidance. I added a migration, too

db/migrate/20180801150624_add_test_track_app_and_app_id.rb
```diff
@@ -0,0 +1,13 @@
+class AddTestTrackAppAndAppId < ActiveRecord::Migration[5.0]
+  class App < ActiveRecord::Base; end
+  class IdentifierType < ActiveRecord::Base; end
+  def up
```
samandmoore: nit: space between this line and the class definitions above it?

samandmoore:
domainlgtm! nice

db/migrate/20180801150624_add_test_track_app_and_app_id.rb
effron: 👍

PR #83, opened 2018-08-31
Update New Relic gem to support distributed tracing

agirlnamedsophia:
### Summary

> This updates the New Relic gem to support distributed tracing.


samandmoore:
<<domainlgtm

~~not sure if we care to explicitly add a provision to the `Gemfile` that says we want newrelic_rpm >= 5.3 to make sure someone doesn't accidentally regress this to a version that doesn't have DT?~~ i think that maybe i'm being too sensitive to that very tiny risk.

samandmoore:
<<platformlgtm too

PR #89, opened 2018-10-23
Rails 5.1

smudge:
### Summary

Making the changes required to run/boot the app on Rails 5.1.

### Other Information

Had to upgrade the `airbrake` gem for 5.1 compatibility, following the upgrade guide [here](https://github.com/airbrake/airbrake/blob/master/docs/Migration_guide_from_v4_to_v5.md#port). I reworked the config in a way that should work with the old ENV vars (where host and port were supplied separately) as well as the new ENV vars (where host can be supplied alone and can include a port).

/platform @jmileham @samandmoore @aburgel @chrislopresto

screen shot 2018-10-23 at 7 02 47 pm


jmileham:
<< domain lgtm platform lgtm

PR #93, opened 2018-11-28
Adds ddtrace gem and configures DataDog

agirlnamedsophia:
### Summary

This adds support for a DataDog integration. It configures the default process names as well so that they will come through as `test_track`, `test_track-http`, `test_track-active-record`, and `test_track-delayed_job`. 

Datadog will only be configured if the environment variable `DATADOG_ENABLED` is set to `true`.

/platform @jmileham @samandmoore

### DIFF
--- a/config/initializers/datadog.rb
+++ b/config/initializers/datadog.rb
@@ -1,0 +1,12 @@
+require 'ddtrace'
+
+unless Rails.env.in?(%w(development test))
+  service_name = Rails.application.class.parent_name.underscore
+
+  Datadog.configure do |c|
+    c.use :rails, service_name: service_name, distributed_tracing: true
+    c.use :active_record, orm_service_name: "#{service_name}-active_record"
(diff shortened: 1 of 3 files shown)


jmileham:
<< domain lgtm from a testtrack standpoint. This seems like a good turnkey config if people want it, and the gem heft can't be that bad regardless.

jmileham:
(leaving platform for somebody DD-aware)

config/initializers/datadog.rb
```diff
@@ -0,0 +1,14 @@
+if ENV['DATADOG_ENABLED']
```
samandmoore: just to be super explicit, this will make it so that if this variable is set to _any value_ the integration will be activated.

samandmoore:
<<platformlgtm

config/initializers/datadog.rb
smudge: Casting ENV vars to booleans is something I'd love to discover a really compact and sharable solution for. It keeps coming up.

PR #94, opened 2019-03-07
Clean out finished_at in create

jmileham:
### Summary

If you want to reenable a finished split via a migration, it won't actually un-finish it. This fixes that.

/no-platform

### DIFF
--- a/app/models/split_creation.rb
+++ b/app/models/split_creation.rb
@@ -7,6 +7,7 @@ class SplitCreation
   validate :split_must_be_valid
 
   def save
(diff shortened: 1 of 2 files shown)


app/models/split_creation.rb
```diff
@@ -7,6 +7,7 @@ class SplitCreation
   validate :split_must_be_valid
 
   def save
+    split.finished_at = nil
```
coreyja: ❤️  Love when it really is this easy!

PR #95, opened 2019-03-14
Upgrade rails

jmileham:
### Summary

Upgrade to catch latest rails point release with security patches.

/no-platform


samandmoore:
<<domainlgtm

PR #96, opened 2019-03-19
set "Vary": "Origin" header

tekdel:
### Summary

We use Test Track on a few domains: example.org and sub.example.org. At example.org we have a link that redirects a user to sub.example.org. When the user lands onto sub.example.org and clicks back in Chrome Browser, it causes the CORS error at example.org, because the request to Test Track server has been cached with `Origin: sub.example.org`. I hope this link will explain what happens.

https://fetch.spec.whatwg.org/#cors-protocol-and-http-caches


jmileham:
Thanks for the contribution Andrey! We'll dig in soon.

/platform @samandmoore @smudge @chrislopresto

jmileham:
<< domain lgtm platform lgtm

Thanks again Andrey!

PR #98, opened 2019-03-31
Jmileham/soft decisions

jmileham:
### Summary

This is largely an optimization but also sows the seeds for some mobile client extensions we're fixing to do. Instead of materializing decisions by overriding all previous assignments with a bulk assignment, we just leave em and use SQL timestamp comparison to make the most recent of the assignment or decision wins when displaying assignments to a customer.

This pattern of using SQL to enrich assignment results contextually will become useful for varying behavior based on feature-completeness on mobile clients in the field.

### Other Information

Will mark up some interesting decisions below.


db/schema.rb
```diff
@@ -10,167 +10,169 @@
...
+    t.json "registry", null: false
+    t.text "hypothesis"
+    t.text "assignment_criteria"
+    t.text "description"
+    t.string "owner"
+    t.string "location"
+    t.integer "platform"
+    t.boolean "feature_gate", default: false, null: false
+    t.string "decision"
+    t.datetime "decided_at"
```
jmileham: Hidden in all this formatting churn, these are the only two new columns.

spec/controllers/api/v1/visitors_controller_spec.rb
```diff
@@ -63,8 +63,8 @@
         )
       end
 
-      it "only queries once per table (visitor, assignment, and split)" do
-        expect { get :show, params: { id: visitor.id } }.to make_database_queries(count: 3)
+      it "only queries twice (visitors, then assignments joined to splits)" do
+        expect { get :show, params: { id: visitor.id } }.to make_database_queries(count: 2)
```
jmileham: Hey, maybe performance is actually better?

spec/models/split_creation_spec.rb
```diff
@@ -41,6 +41,21 @@
     expect(weather.finished_at).to be_nil
   end
 
+  it 'undecides a decided split' do
```
jmileham: This will be important - to undo a decision you can simply reweight a split now. This will work over a migration or an admin panel reweight.

spec/models/visitor_spec.rb
```diff
@@ -21,4 +21,10 @@
   it "has many unsynced assignments" do
     expect(subject).to have_many(:unsynced_splits)
   end
+
+  it "doesn't include assignments to finished splits" do
+    dead_split = FactoryBot.create(:split, finished_at: Time.zone.now)
+    dead_assignment = FactoryBot.create(:assignment, split: dead_split, visitor: subject)
+    expect(subject.reload.assignments).not_to include(dead_assignment)
+  end
```
jmileham: This is an interesting, we've always included all assignments ever, even for finished splits. This was basically a bug. Now it works correctly.

app/controllers/admin/decisions_controller.rb
```diff
@@ -1,19 +1,19 @@
 class Admin::DecisionsController < AuthenticatedAdminController
   def new
     @split = Split.find params[:split_id]
-    @decision = Decision.new
+    @decision = @split.build_decision
   end
 
   def create
     split = Split.find params[:split_id]
-    CreateDecisionJob.perform_later(split, variant: target_variant, admin: current_admin)
```
jmileham: No need to async this anymore cuz it does no bulk assignments.

app/models/decision.rb
```diff
@@ -1,50 +1,27 @@
 class Decision
   include ActiveModel::Model
 
-  attr_accessor :variant, :split, :admin
```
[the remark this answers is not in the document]
jmileham: Also, the only place we stored the admin was in the bulk assignment associated with the decision, which we no longer create, so there's really no schema to hold this admin right now.

app/models/split.rb
```diff
@@ -62,15 +64,15 @@ def reweight!(weighting_registry)
   end
 
   def assignment_count_for_variant(variant)
-    assignments.where(variant: variant).count(:id)
+    Assignment.from(assignments.for_presentation, :assignments).where(variant: variant).count(:id)
   end
 
   def build_decision(params = {})
-    Decision.new({ split: self }.merge(params))
+    Decision.new(params.merge(split: self))
```
jmileham: This makes split not overrideable, which seems more secure.

app/models/split.rb
```diff
@@ -62,15 +64,15 @@ def reweight!(weighting_registry)
...
   end
 
-  def create_decision!(params = {})
-    build_decision(params).tap(&:save!)
+  def create_decision!(params)
+    build_decision(params).save!
```
jmileham:
```suggestion
    build_decision(params).tap(&:save!)
```
This shouldn't have been modified.

app/models/split.rb
```diff
@@ -62,14 +64,14 @@ def reweight!(weighting_registry)
...
+    Assignment.from(assignments.for_presentation, :assignments).where(variant: variant).count(:id)
   end
 
   def build_decision(params = {})
-    Decision.new({ split: self }.merge(params))
+    Decision.new(params.merge(split: self))
   end
 
-  def create_decision!(params = {})
+  def create_decision!(params)
```
jmileham:
```suggestion
  def create_decision!(params = {})
```
this neither.

spec/models/visitor_spec.rb
jmileham: (this should improve performance for longstanding customers and employees a bit)

app/models/assignment.rb
```diff
@@ -20,6 +22,55 @@ class Assignment < ActiveRecord::Base
...
+
+    def overridden_sql_columns
+      [variant_or_decision_as_variant]
+    end
+
+    def variant_or_decision_as_variant
+      <<~SQL
+        case when
+          splits.decided_at is null
+          or assignments.created_at > splits.decided_at
```
jmileham: Should this be updated_at?

jmileham:
Ok, I've tested this in integration locally and had decisions override assignments to be then overridden by another decision to be overridden by another assignment, etc, etc. All works! I did find a gnarly SQL generation bug that I fixed i the process, and now have a test around it.

app/models/assignment.rb
```diff
@@ -37,26 +35,14 @@ def for_presentation(built_at: nil)
...
-      @select_sql ||= (passthrough_sql_columns + overridden_sql_columns).join(",\n")
-    end
-
-    def passthrough_sql_columns
-      passthrough_columns.map { |n| "assignments.#{connection.quote_column_name(n)}" }
-    end
-
-    def passthrough_columns
-      (column_names - NON_PRESENTATION_COLUMNS - %w(variant))
+      Assignment.from(joins(:split).select(select_columns), :assignments).joins(:split).readonly
```
jmileham: This (making the assignments query into a subselect) turned out to be important so that ActiveRecord doesn't assume that it can go straight to the variant column as a source of truth when doing joins. It was wiping out my calculated variant value.

I also refactored for simplicity a little bit in the process.

jmileham:
Realizing that capturing decision in splits is redundant, will just lean into hiding assignments that lose to decisions and leaning into the split weightings. Feels more honest.

spec/features/admin_split_decision_create_spec.rb
```diff
@@ -36,15 +36,13 @@
...
-    Delayed::Worker.new.work_off
+    expect(split_page).to have_content "Decided"
 
     split_page.load split_id: split.id
     expect(split_page).to be_displayed
 
-    expect(Assignment.where(variant: "hammer_time").count).to eq 0
-    expect(Assignment.where(variant: "touch_this").count).to eq 5
+    expect(split.assignment_count_for_variant(:hammer_time)).to eq 0
+    expect(split.assignment_count_for_variant(:touch_this)).to eq 0
```
jmileham: Now assignments just hide themselves completely when a decision overrides them.

app/controllers/admin/decisions_controller.rb
```diff
@@ -1,19 +1,19 @@
...
-    flash[:success] = "Queued decision to reassign all visitors to #{target_variant}"
+    split.create_decision!(decision_params)
```
smudge: Something about having surface area on a model for a _different_ model's persistence operations feels strange (even though I know that Rails autogenerates the same `create_` methods for AR associations). If the method didn't already exist, I'd suggest using `@split.build_decision` directly (and calling `#save!` on the result).

app/models/assignment.rb
```diff
@@ -20,6 +20,41 @@ class Assignment < ActiveRecord::Base
 
   delegate :name, to: :split, prefix: true
 
+  class << self
+    def to_hash
+      Hash[all.includes(:split).map { |a| [a.split.name.to_sym, a.variant.to_sym] }]
+    end
```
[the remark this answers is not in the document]
smudge: > I needed a private class method

To be clear, this is one that you're adding in a follow-up PR?

app/models/split.rb
```diff
@@ -2,7 +2,7 @@ class Split < ActiveRecord::Base
   belongs_to :owner_app, required: true, class_name: "App", inverse_of: :splits
 
   has_many :previous_split_registries, dependent: :nullify
-  has_many :assignments, dependent: :nullify
+  has_many :assignments, -> { for_presentation }, dependent: :nullify, inverse_of: :split
```
smudge: For reference, I've never encountered a good way to pass arguments through from associations to scopes, so the `.assignments` association will probably never be able to dynamically adjust to anything other than the current state of the DB. This might make it tricky in certain controllers where you need to do `assignments.for_presentation(some_args)`, because it'll already have `for_presentation` (no args) baked into it

app/models/split.rb
smudge: Sometimes the key order matters, but that doesn't seem to be the case here, nor would this be (as you said) the most secure way to solve the ordering problem.

app/models/assignment.rb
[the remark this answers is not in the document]
jmileham: I actually had several private class methods and then refactored them all away in this PR, but yeah, they're back in the next one. :)

app/models/split.rb
jmileham: Yeah, we'll see how to do this. I might make a pseudo-association instead. Its usage is read-only anyway.

jmileham: I think the correct factoring for all this will become clearer as I need to wire it into the mobile-specific controllers and provide it an app_build.

app/models/assignment.rb
```diff
@@ -20,6 +20,40 @@ class Assignment < ActiveRecord::Base
...
+
+    def for_presentation(built_at: nil)
+      q = presentation_query
+
+      q = built_at.nil? ? q.where("splits.finished_at is null") : q.where("splits.finished_at > ?", built_at)
+
+      q
```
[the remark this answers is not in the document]
smudge: Any reason not to make this a `scope`? I'm also wondering if it is worth trying to deconstruct into subparts (especially in anticipating additional variants) or if it is better to leave it in one big colocated query method the way you have it:

```ruby
scope :for_presentation, ->(built_at: nil) { undecided.unfinished(as_of: built_at) }

# assignment.rb
scope :undecided, -> { joins(:split).where('splits.decided_at IS NULL OR assignments.updated_at > splits.decided_at') }
scope :unfinished(as_of:) -> { joins(:split).merge(Split.unfinished(as_of: as_of)) }

# split.rb
scope :unfinished, ->(as_of:) { as_of ? where('splits.finished_at > ?', as_of) : where(finished_at: nil) }
```

(reactions to that?)

app/controllers/admin/decisions_controller.rb
jmileham: yeah, I'm not married to it, it feels a little railsy on the outside, but it's not doing anything for keeping split from becoming a god model.

app/models/assignment.rb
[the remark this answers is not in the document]
jmileham: yeah, this looks good. I'm inclined to walk toward this in subsequent PRs, if that makes sense, 'cuz I've already got another working PR based on this factoring. I like the composable scopes thing, effectively they're not great developer experience on their own but as long as I expose macro scopes that do everything in a coordinated way that's fine.

Stylistically, I generally have preferred class methods whenever a scope-like thing takes args because the syntax gets a little nuts. I think one of the rails core folks has a blog post about how that's his preference as well. Do you have a reason you prefer scopes even with args?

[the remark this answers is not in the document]
jmileham: Also, if I keep the composable-and-yet-not-ideal-developer-experience scopes as private class methods, that's probably a win too.

app/models/decision.rb
```diff
@@ -1,50 +1,23 @@
...
-
-  def bulk_assignment
-    @bulk_assignment ||= BulkAssignment.new(admin: admin, split: split, reason: "Decision", variant: variant)
-  end
+  private
 
-  def variant_belongs_to_split
-    errors.add(:variant, "#{variant} is not a valid variant for split") unless split.has_variant? variant
+  def build_split_creation
+    split.build_split_creation(
```
smudge: Normally the "creation" wording is fine if we're saying "i intend to create if it doesn't already exist, otherwise update it." But here what we're saying is "I know it already exists and I want to update it" -- I've started calling these classes `Upsert`s instead of `Creation`s, but I guess we gotta live with the current naming for this PR at least. 🤷

app/models/split.rb
smudge: It might also be possible to assign transient state to the split instance and then read that from the association, but that seems like all kinds of magic.

spec/models/visitor_spec.rb
smudge: Nice! 👍

app/models/decision.rb
jmileham: I mean, we don't *have* to, but I am trying to keep the diff size down on this guy, but I like your thinking on that naming. I'll try to get that happening over time.

app/models/split.rb
jmileham: yep. i will try desperately to avoid. :)

app/models/assignment.rb
jmileham: (this went away)

app/models/assignment.rb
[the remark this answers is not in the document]
smudge: My sole reason for preferring `scope` is that it tells me exactly what it returns, so I don't have to read the guts and do the variable-hopping to determine that oh, it in fact returns an ActiveRecord relation and is basically a `scope` in sheep's clothing.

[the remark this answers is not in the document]
smudge: But I totally buy that single-use monster scopes work better on net as class methods. We can defer this to later PRs when we really know what we're dealing with here.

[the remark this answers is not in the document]
smudge: Also, private scopes should be a thing.

app/models/decision.rb
```diff
@@ -1,50 +1,23 @@
...
-
-  def weight_split_fully_to_variant
-    split.reweight!(variant => 100)
-  end
-
-  def reassign_existing_assignments
-    bulk_assignment.save!
-    existing_assignments.find_in_batches do |subset|
-      BulkReassignment.create!(assignments: subset, bulk_assignment: bulk_assignment)
+    raise "Variant must be present in the split" unless split.has_variant?(variant)
```
smudge: Assuming the choice to inline the `variant_belongs_to_split` validation is because it's the only meaningful validation now?

app/models/decision.rb
```diff
@@ -1,50 +1,23 @@
...
+    build_split_creation.save!.tap do
+      split.reload
```
smudge: Feels like knowing to reload could be `SplitCreation`'s responsibility in the upsert case.

app/models/decision.rb
jmileham: It was mostly the fact that I churned this file out of existence and back again. I'll put it back. It'd make it possible to use regular save in the controller too.

app/models/decision.rb
jmileham: but then it'd need to know it was an update of an instance, which it doesn't. Unless I add a `#create_split_creation!` method to `Split`, which maybe isn't worth it?

app/models/decision.rb
jmileham: Oh, actually it was to simplify the error bubbling - because I didn't need non-raisey save of decision, I figured I could skip all that and just make it blow up. Decision creations are never done by unwashed peeps - it'll either be admins or migration clients who should know better. That said, I may need real validations later to make the migration client spit actionable error messages. So I might be back to square one before I know it.

app/models/decision.rb
smudge: It could definitely look at `split.new_record?` before calling save on it. Whether it _should_ gets a little hand-wavey, but typically I see an explicit `reload` as a state-related smell and I'd want to push it as close to the persistence operation as possible (if not fix whatever bug isn't properly setting state during persistence).

jmileham: The problem is that the instance in the upsert is always new, right? I don't get to reload the split instance that the caller has reference to from the split_creation no matter what, unless some identity map stuff has been completed that I'm unaware of in modern rails?

jmileham: really what's happening here is we're favoring reuse of simple upsert codepath over enough decoupling to make reloading unnecessary. I think I might rip out `Split#build_split_creation` in favor of a `Split.reconfigure!` method that secretly encapsulates `SplitCreation` and does the reloading nonsense. I can do that in this PR and see how it looks.

smudge: Oh, I see. Yes, that is correct, it's not gonna mutate `split` because the way `split.build_a_thing_that_creates_me_even_though_i_exist` (do you like my naming suggestion :trollface:) were smart enough to short-circuit the "find_or_initialize" and just hand it `self`. Which doesn't do.

So you're totally right, this needs an explicit reload 👍

app/models/decision.rb
smudge: Yep, totally fair. BTW if you were gonna go back to `raise errors.full_messages.to_sentence unless valid?` I would just change it to `validate!`.

spec/features/admin_split_decision_create_spec.rb
smudge: Do we not show the count somewhere on the page? Might be nice to avoid reaching straight to the model code if something on `split_page` can tell us what we need.

spec/models/split_creation_spec.rb
smudge: As long as this won't allow us to change the semantics of an existing split thinking we're creating a new one. Like, if we decide a split and it's really only known to old mobile clients, and then in the future we reopen the split and accidentally enable something we shouldn't on those same old mobile clients...

spec/models/assignment_spec.rb
```diff
@@ -90,4 +90,45 @@
...
+
+    it "doesn't return assignments for splits finished before built_at if provided" do
+      finished_split = FactoryBot.create(:split, name: "finished_split", finished_at: Time.zone.now)
+      FactoryBot.create(:assignment, split: finished_split)
+      built_at = 5.minutes.from_now
+
+      expect(described_class.for_presentation(built_at: built_at).map(&:split_name)).not_to include("finished_split")
+    end
+
+    it "returns assignments finished after built_at if provided" do
```
smudge: `assignments for splits finished after built_at`?

spec/models/assignment_spec.rb
```diff
@@ -90,4 +90,45 @@
...
+      expect(described_class.for_presentation).to be_empty
+    end
+
+    it "returns the assignment if newer than the split decision" do
+      split.create_decision!(variant: :touch_this)
+      FactoryBot.create(:assignment, split: split, variant: :hammer_time)
+
+      expect(described_class.for_presentation.first.variant).to eq("hammer_time")
+    end
+  end
```
smudge: What does it take for a split to be both decided and finished? Do we care about:

- Assignments older than decision for splits finished after built_at
- Assignments newer than decision for splits finished after built_at
- Assignments newer than decision for splits finished _before_ built_at
- etc?

I know the query construction makes those less likely to be wrong, but I could see someone refactoring it and accidentally missing some kind of interference between the two parts of the query.

spec/features/admin_split_decision_create_spec.rb
jmileham: Yeah, that's a fair shout, but this is super Not My Test :) There's a lot of work I'd do to the admin screens and tests given the chance, but I feel like that might be a bridge too far for this PR.

smudge:
<< domain LGTM - you've answered most of my questions, and I think the interesting decisions are gonna come in some subsequent PRs, so I don't want to block you on anything here.

spec/models/split_creation_spec.rb
jmileham: Yeah, I'm not crazy worried about that given our conventions around quarter-specific split naming. Also I'm going to do more in subsequent PRs (around supporting a new migration runner API) to shore up app ownership of splits so nobody can mutate another app's splits, which should keep spooky action at a distance down.

spec/models/assignment_spec.rb
jmileham: yup.

```suggestion
    it "returns assignments for splits finished after built_at if provided" do
```

spec/models/assignment_spec.rb
jmileham: Yeah, could add those. I don't know if we're gonna get the full cross product though, especially once we add more variables like feature completions and remote kills. Do you have strong feels that any of these are important coverage we're missing?

smudge: One strategy for testing scopes with many input combinations is to lean more on the factories to set things up in a variety of conditions (like, "i expect the query against [ABCDE] to return [ABD]"). I don't feel super strongly here, but it might be nice to have one additional assignment that falls into two groups just to acknowledge the possibility for future devs.

jmileham: I'm hesitant to add more tests now since the composition of this logic is likely to change in ways that would allow me to get the confidence that things will behave correctly in integration based on unit testing the smaller scopes. I'll take notes on the open Qs from this PR as a tickler file for the overall arc here, though, and sufficient coverage of condition combinations is a thing I'll re-review.

app/controllers/admin/decisions_controller.rb
jmileham: One nice thing in retrospect is that `create_decision!` reloads split based on the decision, so it kind of is related.

app/models/decision.rb
jmileham: so I pushed up a change to start hiding the underlying model naming by calling this `build_change` and making a `reconfigure!` method that does that plus self.reload. This is feeling on a good track. I'll rename the creation type to upsert in a subsequent PR.

jmileham:
bump! I started heading in the direction of renaming the SplitCreation model and have the following additional notes I'll work through as I go:

* Decompose into smaller scopes over time
    * Maybe use scopes for real because they obviously return relations? But private class methods are nice too.
* Make sure app ownership is strong in split modification APIs
* Figure out a pattern for presenting assignments in a way that's safe but allows argument passing.
    * Ideally avoid transient state in split/visitor instance
* SplitCreation -> SplitUpsert
* Figure out testing story that establishes confidence that condition combinations work correctly together.
    * Either more tests or unit testing of smaller scopes and demonstration their participation in higher-level scopes, or a few representative tests of split/app_build condition combinations.

Would love platform review too - Sam is out of pocket for the moment, so he won't be sad if you take it.

samandmoore:
> But private class methods are nice too.

I don't think you can dispatch a private class method against a Relation. We ran into this somewhere else. I guess you could `send` it though.

samandmoore:
Yea. Feel free to take platform. I'm reading along when I can but am not super available the next few days 😔

jmileham:
Yeah, I'm going to real scopes in my refactor anyway. It is looking aight. Multiline stabby lambads aren't crazy looking.

smudge:
<< platform LGTM on everything in this PR so far -- will take a look at the subsequent PR when I get the chance. (LMK when the scope refactor is ready for eyes)

PR #100, opened 2019-04-02
Feature completion

jmileham:
### Summary

This adds a FeatureCompletion model and wires it into split visibility.

Notably this doesn't actually affect production behavior, and it's missing a piece before it would be safe. We also need to wire up modifications to the weighting registry for display to the same app (basically overriding displayed weightings to 100% `false` for feature gates on app versions that aren't in the feature-complete set).

Will mark up below.


app/models/app_version.rb
```diff
@@ -0,0 +1,48 @@
+class AppVersion
+  include Comparable
+
+  DECIMAL_INTEGER = /(?:0|[1-9]\d*)/
```
jmileham: This regexp (used within the `VERSION_REGEX`) disallows ruby octal literal syntax. Trying to make sure this thing doesn't have surprising behavior.

app/models/app_version.rb
```diff
@@ -0,0 +1,48 @@
...
+  def to_a
+    @to_a ||= version_number.split('.').map { |n| Integer(n) }
+  end
+
+  def to_s
+    version_number
+  end
+
+  def to_pg_array
+    "{#{to_a.join(',')}}"
```
jmileham: AppVersions are comparable both in PG query-land and in ruby-land thanks to a neat property of arrays on both. An array of integers corresponding to a de-dotted version number collates correctly in ruby and PG.

app/models/app_version.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def inspect
+    "#<AppVersion: #{self}>"
+  end
+
+  def <=>(other)
+    to_a <=> other.to_a
```
jmileham: `Comparable` works nicely for us here. But one semi awkward effect is that you can compare this to anything that has a to_a method, but I just don't think that matters right now.

app/models/assignment.rb
```diff
@@ -13,23 +13,40 @@ class Assignment < ActiveRecord::Base
...
-    end
+  scope :unsynced_to_mixpanel, -> { where("mixpanel_result = 'failure' OR mixpanel_result IS NULL") }
 
-    def for_presentation(built_at: nil)
-      q = joins(:split)
-        .where("splits.decided_at is null or assignments.updated_at > splits.decided_at")
-      q = built_at.nil? ? q.where("splits.finished_at is null") : q.where("splits.finished_at > ?", built_at)
-      q
-    end
+  scope :for_presentation, ->(app_build: nil) do
```
jmileham: Rather than pass a bag of disjunct args at this scope, I realized a type representing an app build was better - you either aren't presenting for a specific app build (and you get the default behavior for Rails/JS web clients), or you are, and you care about version, build date, and which app is in play. (theoretically you could have JS clients in the field at some point, which this version-aware code will enable too).

app/models/feature_completion.rb
```diff
@@ -0,0 +1,9 @@
+class FeatureCompletion < ActiveRecord::Base
+  belongs_to :split
+  belongs_to :app
+
+  validates :split, uniqueness: { scope: :app }
+  validates :split, :app, :version, presence: true
+
+  attribute :version, :app_version
```
jmileham: I used the Rails 5 Attributes API to make this type conversion work. It's great and will come in handy for auto-casting string representations of version numbers coming in via RESTful APIs later.

app/models/split.rb
```diff
@@ -19,7 +19,9 @@ class Split < ActiveRecord::Base
 
   before_validation :cast_registry
 
-  scope :active, -> { where(finished_at: nil) }
+  scope :active, ->(as_of: nil) do
+    as_of ? where('splits.finished_at is null or splits.finished_at > ?', as_of) : where(finished_at: nil)
+  end
```
jmileham: Realized I already had the start of this scope, so I refactored it to take an optional argument, and use it to drive out the composable scope refactor you suggested on #98, @smudge.

app/types/app_version_type.rb
```diff
@@ -0,0 +1,15 @@
...
+  rescue StandardError
+    nil
+  end
+
+  def serialize(value)
+    value.to_pg_array if value
+  end
+
+  def deserialize(value)
+    AppVersion.from_pg_array(value) if value
```
jmileham: Again, the customer-facing casting from string is different than the database-facing casting-to-PG-array. It all works nicely, though, and we can just use DB comparison operators on pairs of arrays-representing-version-numbers.

spec/models/assignment_spec.rb
```diff
@@ -92,43 +92,158 @@
   end
 
   describe ".for_presentation" do
-    let(:split) { FactoryBot.create(:split) }
+    it "filters for one reason of underlying scopes (decision overrides)" do
```
jmileham: @smudge this is attempting to establish some confidence that the composed scopes still reflect the behavior of the underlying scopes `and`ed-together without boiling the combinatorial ocean, because if they were `or`ed-together, this filtering wouldn't work.

spec/models/assignment_spec.rb
```diff
@@ -92,43 +92,158 @@
...
+    it "combines excluding_decision_overrides and for_active_splits (argless) with no app_build" do
+      allow(described_class).to receive(:excluding_decision_overrides).and_call_original
+      allow(described_class).to receive(:for_active_splits).and_call_original
+      expect(described_class).not_to receive(:for_app_build)
+
+      described_class.for_presentation
 
-      expect(described_class.for_presentation.map(&:split_name)).not_to include("finished_split")
+      expect(described_class).to have_received(:excluding_decision_overrides)
+      expect(described_class).to have_received(:for_active_splits).with(no_args)
```
jmileham: Otherwise, we just assert on what scopes the combined scopes call, while at least ensuring that they don't raise exceptions. Seems like the scope-equivalent of spying on your OO collaborators, which I do a fair bit in model specs in order to assert that the plumbing works in integration while doing the in-depth testing on units.

spec/models/split_spec.rb
```diff
@@ -166,4 +166,37 @@
...
+      expect(described_class.active(as_of: Time.zone.now)).to include(split)
+    end
+
+    it "returns splits finished after as_of:" do
+      split = FactoryBot.create(:split, finished_at: Time.zone.now)
+
+      expect(described_class.active(as_of: 1.minute.ago)).to include(split)
+    end
+
+    it "doesn't return splits finished at simultaneously with as_of:" do
```
jmileham:
```suggestion
    it "doesn't return splits finished simultaneously with as_of:" do
```

app/models/assignment.rb
jmileham: BTW, my current hunch is that I'm going to eliminate the visitor-to-assignments association and replace it with a class method like so:

```ruby
def self.assignments(params = {})
  Assignment.where(split: self).for_presentation(params)
end
```

I just don't need a fully-loaded association. I might need association-cache aware memoization though, so that might get sticky.

Anyway, that's still for later.

spec/models/assignment_spec.rb
```diff
@@ -92,43 +92,158 @@
...
     end
 
-    it "doesn't return assignments for splits finished before built_at if provided" do
-      finished_split = FactoryBot.create(:split, name: "finished_split", finished_at: Time.zone.now)
-      FactoryBot.create(:assignment, split: finished_split)
-      built_at = 5.minutes.from_now
+    it "combines excluding_decision_overrides and for_app_build when app_build is provided" do
+      app = FactoryBot.create(:app)
```
jmileham:
```suggestion
      app = FactoryBot.build_stubbed(:app)
```

spec/models/assignment_spec.rb
```diff
@@ -92,43 +92,158 @@
...
     end
+  end
 
-    it "returns assignments for splits finished after built_at if provided" do
-      finished_split = FactoryBot.create(:split, name: "finished_split", finished_at: Time.zone.now)
-      FactoryBot.create(:assignment, split: finished_split)
-      built_at = 5.minutes.ago
+  describe ".for_app_build" do
+    it "combines for_active_splits and excluding_incomplete_features, forwarding args from app_build" do
+      app = FactoryBot.create(:app)
```
jmileham:
```suggestion
      app = FactoryBot.build_stubbed(:app)
```

app/models/assignment.rb
```diff
@@ -13,23 +13,40 @@ class Assignment < ActiveRecord::Base
...
-
   normalize_attributes :mixpanel_result
 
   delegate :name, to: :split, prefix: true
 
-  class << self
-    def to_hash
-      Hash[all.includes(:split).map { |a| [a.split.name.to_sym, a.variant.to_sym] }]
-    end
+  scope :unsynced_to_mixpanel, -> { where("mixpanel_result = 'failure' OR mixpanel_result IS NULL") }
```
smudge: I don't think this gets used anymore, does it? It feeds all the way back up to `unsynced_splits`, which I don't see any non-spec references to.

app/models/app_version.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def initialize(version_number)
+    raise "version_number must be a string, integer, or AppVersion" unless [String, Integer, AppVersion].any? do |t|
+      version_number.is_a?(t)
+    end
+    version_number = version_number.to_s
+    raise "version_number must be present" if version_number.blank?
+    raise "version_number is too long" if version_number.length > MAX_LENGTH
+    raise "version_number does not conform to format" unless version_number.match(VERSION_REGEX)
```
smudge: So if someone changes/tightens any of these constraints after the fact, we'll `rescue` below in `FeatureCompletion` and get `nil` on the version. What are the downstream effects of that, and were does/should the nil-handling occur? (I'm having trouble tracing all of the entry points -- like is `define_build` actually used in a non-spec context?)

smudge: I'm mostly worried about a code bug / developer error resulting in something being shown that shouldn't because the version-handling breaks down somewhere.

app/models/app_version.rb
smudge: Yeah seems fine.

app/models/assignment.rb
jmileham: k, i'll double check and kill it if so.

app/models/app_version.rb
jmileham: So we've got a not-null constraint and presence validation on the FeatureCompletions table which will make sure that feature completions will always persist with well-formed versions. define_build will not rescue the exception because it doesn't run through the attributes API, so you'll get a hard foreground error. If I need to elevate that to validations in order to put better client errors in place, I can, but I suspect I can require the clients to validate the version format such that 500 errors are acceptable if a client pitches us a bad version.

jmileham: And yeah, the entrypoints will be twofold:

* Migrations from client side will create/mutate feature_completions
* Mobile clients will make visitor requests with an app_name, version_number and built_at in the payload which will be used to `define_build` and then the resulting app_build will be passed into the `for_presentation` scope for display (plus whatever equivalent thing needs doing to modify the split weighting registry to respect feature incompletions)

app/models/assignment.rb
jmileham: kilt

app/models/app_version.rb
smudge: I think it's the latter that might break when you go from `request` -> `version` (nil) -> `to_pg_array` (in `for_presentation`) -- again only if the data at rest used to be valid and no longer upon decoding / casting it to an `AppVersion`.

smudge: It's not a huge risk, nor a direct result of anything in the current code -- but things do drift over time in surprising ways, and there is nothing stopping someone from coming in and "correcting" these `raise` calls in a way that invalidates existing production data.

app/models/assignment.rb
```diff
@@ -13,23 +13,38 @@ class Assignment < ActiveRecord::Base
...
+  end
+
+  scope :excluding_incomplete_features, ->(app_id:, version:) do
+    joins(:split).where(<<~SQL, app_id, AppVersion.new(version).to_pg_array)
+      splits.feature_gate = false
+      or exists (
+        select 1 from feature_completions fc
+        where fc.app_id = ?
+          and fc.split_id = assignments.split_id
+          and fc.version <= ?
```
smudge: So, I get that this is where the very elegant magic happens -- and ultimately why versions are treated as value objects (so that they can cast themselves to and from pg arrays).

I'm curious about how/whether this can be indexed. (Trying to read the pg docs, I can't tell if a B-tree index explicitly doesn't work for array columns, or if it works but not for every array operation.)

The alternative, in my mind, would be to store three separate integers, which can most certainly be indexed but would, I imagine, result in a more complex query.

app/models/app_version.rb
smudge: Oh wait, ignore me. I see now that it's `define_build` that will blow up with `AppVersion.new` -- already-persisted feature completions won't break that comparison because we're _not_ loading them up again at that point and just reading straight from the table. 🤦‍♂️ 

I got it, feel free to ignore me. I'll wrap my head around all of this eventually. 😅

app/models/assignment.rb
jmileham: I’m not using version to restrict my working set that significantly, so indexing is probably not worth it - the set of assignments I have as a visitor is the highest cardinality index.

app/models/app_version.rb
jmileham: Deserializing probably would blow up in surprising ways, but you’d have already fixed my tests to make a change to a more restrictive format. I based my format on an Apple technical publication, and to my eye there’s no air gap between what Apple will allow and what we will allow in either direction, so I’d be surprised to find this getting more restrictive, if anything less to support 3-dot versions and alpha/beta/rc. Doing the textual stuff would require a migration to allow the db column to store text so that’d be the opportunity to true up platter representation with new format.

app/models/app_version.rb
```diff
@@ -0,0 +1,48 @@
+class AppVersion
+  include Comparable
+
+  DECIMAL_INTEGER = /(?:0|[1-9]\d*)/
+  VERSION_REGEX = /\A(?:#{DECIMAL_INTEGER}\.){0,2}#{DECIMAL_INTEGER}\z/ # iOS rules currently, but can be relaxed from here
```
smudge: So iOS seems to be explicitly 3 integers always (at least from [reading their docs](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-111349)). I could be wrong on that.

Seems like android is the one that might need it relaxed. [Their docs](https://developer.android.com/studio/publish/versioning) make it seem like `versionName` is _intended_ to follow (major).(minor).(patch) formatting, but then their example doesn't.

Anyway, what's the story for Android support here? Especially if we leave the land of well-formed integers? Can we still force funkier version formats into a pg array comparison?

jmileham: See the tests - Android will proffer the monotonic integer build thingie instead of a version number.

jmileham: Also agreed - this supports up to 3 integers like the Apple spec because the Apple spec says so. :) am I missing something?

app/models/assignment.rb
```diff
@@ -13,23 +13,38 @@ class Assignment < ActiveRecord::Base
...
+  scope :excluding_decision_overrides, -> do
+    joins(:split).where('splits.decided_at is null or assignments.updated_at > splits.decided_at')
+  end
+
+  scope :for_active_splits, ->(as_of: nil) do
+    joins(:split).merge(Split.active(as_of: as_of))
+  end
+
+  scope :excluding_incomplete_features, ->(app_id:, version:) do
+    joins(:split).where(<<~SQL, app_id, AppVersion.new(version).to_pg_array)
```
smudge: In theory this isn't actually assignment-specific, so you could move it up to `Split` alongside `active`, but at that point I guess it's just kind of spreading out the concepts:

```ruby
joins(:split).merge(Split.for_complete_features(app_id: app_id, version: version))
```

So happy leaving this as-is.

app/types/app_version_type.rb
```diff
@@ -0,0 +1,15 @@
+class AppVersionType < ActiveRecord::Type::Value
+  def cast(value)
+    AppVersion.new(value) unless value.nil?
+  rescue StandardError
+    nil
+  end
+
+  def serialize(value)
+    value.to_pg_array if value
```
smudge: Rubocop didn't make you use safe navigation?

spec/models/assignment_spec.rb
```diff
@@ -80,55 +80,159 @@
...
+    end
+
+    it "doesn't return assignments for features completed after the provided version" do
+      split = FactoryBot.create(:split, feature_gate: true)
+      assignment = FactoryBot.create(:assignment, split: split)
+      feature_completion = FactoryBot.create(:feature_completion, split: split, version: "1.0.0")
+
+      expect(
+        described_class.excluding_incomplete_features(
+          app_id: feature_completion.app.id,
```
smudge: Worth adding a case for when the app ID doesn't match?

app/models/app_version.rb
smudge: I guess my question was more how Android enforces constraints on top-level `versionName` values. Like, is there something stopping me from releasing versions like `1.2-beta1` and `1.2-beta6`?

smudge: (In their example they have a `1.1-demo` as one of the "flavors" of release, wasn't sure if the formatting gets stricter outside of the flavors.)

app/models/assignment.rb
smudge: Makes sense 👍

app/models/assignment.rb
jmileham: I may refactor that direction when making split weighting registries behave, or maybe that’ll work better in ruby, tbd.

app/types/app_version_type.rb
jmileham: TT is not on the latest house stuff prolly.

app/models/app_version.rb
jmileham: We won’t touch that number, Android has a second secret build number integer they require each build to sport so they don’t even bother to have collatable version numbers for sorting in the store. we’ll ingest the integer.

jmileham: I’d only be interested in adding text support for a platform like a desktop app or something that had a richer convention around prerelease versions but still had a formal collation algorithm.

smudge:
Alright, so I think the README might need a bit of a terminology injection at this point, to include "feature completions" and also to make it clear how deciding a split (and setting "decided_at") compares to removing a split (which soft-deletes using "finished_at"), especially w.r.t. how mobile clients interpret all of that (since older clients can effectively circumvent both split decisions and soft deletions).

Happy to do that here, or wait until you finish wiring things up.

Also curious if you have any thoughts -- I am sort of questioning the "completion" terminology, to avoid confusion with the other two termination-sounding states/column values ("decided_at" and "finished_at"). What about "feature launch", to imply that it's the _start_ of the feature's availability, not the _end_. It's silly but I legit had about 5 minutes of confusion trying to figure out the use case for telling apps when a feature split has been "completed" (i.e. finished?).

smudge:
<< platform LGTM && domain TAFN

smudge:
Did another pass! Thanks for talking me through to arriving at the missing hyphen in my mental model.

domain LGTM aside from the one spec case.

jmileham:
Note to self: there's a future feature that will add to/modify this behavior. Once we have "hard" assignments (future PR), hard assignments more recent than a relevant feature_completion will win over those feature completions.

smudge:
Re-upping the LGTM on the two recent commits 👍

app/models/app_version.rb
jmileham: Also, Here's the apple technical note I referenced - they are apparently down with one-to-three integers, not just exactly three, so for the time being, I think I have algo parity with Apple (though I'm open to relaxing it further to support other platforms - it just so happened that Android `versionCode`s are a subset of the Apple spec): https://developer.apple.com/library/archive/technotes/tn2420/_index.html#//apple_ref/doc/uid/DTS40016603-CH1-NUMBERING_CONVENTIONS-EXAMPLES

I'm going to link this tech note into the comments of the REGEX too.

PR #102, opened 2019-04-03
Upgrade rubocop

jmileham:
### Summary

To answer your question, it didn't [force safe navigation](https://github.com/Betterment/test_track/pull/100#discussion_r271521460) after all. There were a few classes of thing in here:

* A bunch of autocorrects, took em, tests passed, yolo (no I looked, they were uncontroversial).
* Did a raisey create instead of an assertion in a before block for test setup
* Disabled rubocop inline where it was misinterpreting a factory for an RSpec, this is just not worth fixing in rubocop given the tiny impact surface area (a factory for a model with a `context` attribute living in `spec/**`).
* generated a TODO for 12 instances of stubbing the subject under test. I'm on board with this, but I don't think it's super important that I address these right now, which is why I'm leaving them in the TODO as debt we should think about fixing, but the right solution will often be object decomposition, so not the speediest.

/domain @smudge 
/no-platform


spec/factories/assignments.rb
```diff
@@ -2,7 +2,8 @@
   factory :assignment do
     visitor
     split
-    context 'default_context'
+    # rubocop misinterprets below as an RSpec context:
+    context { 'default_context' } # rubocop:disable RSpec/EmptyExampleGroup, RSpec/MissingExampleGroupArgument
```
smudge: Someone should tell rubocop that context matters 😉

smudge:
<< domain LGTM

PR #103, opened 2019-04-04
Knockout incomplete feature weightings

jmileham:
### Summary

This knocks out the split registry to 100% false for a given split in response to feature incompleteness on the current client app build. In conjunction with #100, it means that folks who are using a version of a client that is not feature complete for a given gate will not see any explicit assignments for that gate (PR #100's doing), and instead will see weightings of 100% false (this PR's doing) that cause the client to disable the feature.

There will be a future PR to add "hard" assignments that will win over these knockouts. I expect the weightings to remain knocked out in that case, we'll just ship the hard assignments to the client to layer over top.

/platform @smudge @samandmoore

### DIFF
--- a/app/models/split.rb
+++ b/app/models/split.rb
@@ -131,10 +131,10 @@ class Split < ActiveRecord::Base
 
   def knock_out_weightings(registry_hash, to: "false")
     target_variant = to.to_s
-    knock_out_hash(registry_hash, to: target_variant) || log_knockout_error(target_variant) && registry_hash
+    knock_out_weightings_if_possible(registry_hash, to: target_variant) || log_knockout_error(target_variant) && registry_hash
   end
 
-  def knock_out_hash(registry_hash, to:)
+  def knock_out_weightings_if_possible(registry_hash, to:)
     found_key = false
     knocked_out_registry = registry_hash.each_with_object({}) do |(k, _), h|
       h[k] = (k.to_s == to ? (found_key = true && 100) : 0)
--- a/app/models/split.rb
+++ b/app/models/split.rb
@@ -120,7 +120,7 @@ class Split < ActiveRecord::Base
   end
 
   def registry
-    if try(:feature_incomplete?)
+    if try(:feature_incomplete?) # This is a virtual attribute provided by the with_feature_incomplete_knockouts_for scope
       knock_out_weightings(super)
     else
       super
--- a/app/models/split.rb
+++ b/app/models/split.rb
@@ -34,15 +34,15 @@ class Split < ActiveRecord::Base
     as_of ? where('splits.finished_at is null or splits.finished_at > ?', as_of) : where(finished_at: nil)
   end
 
-  scope :with_feature_completeness_for, ->(app_build) do
+  scope :with_feature_incomplete_knockouts_for, ->(app_build) do
     previous_selects = all.arel.projections
     except(:select)
       .select(
         previous_selects,
         Arel::SelectManager.new
-          .where(arel_excluding_incomplete_features_for(app_build))
+      .where(arel_excluding_incomplete_features_for(app_build).not)
           .exists
-          .as('feature_complete')
+          .as('feature_incomplete')
       )
       .readonly
   end
@@ -120,9 +120,9 @@ class Split < ActiveRecord::Base
   end
 
   def registry
-    if respond_to?(:feature_complete?) && !feature_complete?
-      super.each_with_object({}) do |(k, v), h|
-        h[k] = (v == "false" ? 100 : 0)
+    if try(:feature_incomplete?)
+      super.each_with_object({}) do |(k, _), h|
+        h[k] = (k.to_s == "false" ? 100 : 0)
       end
     else
       super
--- a/app/models/split.rb
+++ b/app/models/split.rb
@@ -27,7 +27,7 @@ class Split < ActiveRecord::Base
 
   scope :for_app_build, ->(app_build) do
     active(as_of: app_build.built_at)
-      .with_feature_completeness(app_build)
+      .with_feature_incomplete_knockouts_for(app_build)
   end
 
   scope :active, ->(as_of: nil) do
--- a/app/models/split.rb
+++ b/app/models/split.rb
@@ -17,6 +17,8 @@ class Split < ActiveRecord::Base
   validate :registry_weights_must_be_integers
   validate :registry_must_have_winning_variant_if_decided
 
+  enum platform: %i(mobile desktop)
+
   before_validation :cast_registry
 
   scope :for_presentation, ->(app_build: nil) do
@@ -49,14 +51,16 @@ class Split < ActiveRecord::Base
     where(arel_excluding_incomplete_features_for(app_build))
   end
 
-  def self.arel_excluding_incomplete_features_for(app_build)
-    Arel::Nodes::Or.new(
-      arel_table[:feature_gate].eq(false),
-      AppFeatureCompletion.select(1).satisfied_by(app_build).arel.exists
-    )
-  end
+  class << self
+    private
 
-  enum platform: %i(mobile desktop)
+    def arel_excluding_incomplete_features_for(app_build)
+      Arel::Nodes::Or.new(
+        arel_table[:feature_gate].eq(false),
+        AppFeatureCompletion.select(1).satisfied_by(app_build).arel.exists
+      )
+    end
+  end
 
   def detail
     @detail ||= SplitDetail.new(split: self)
--- a/app/models/split.rb
+++ b/app/models/split.rb
@@ -19,8 +19,32 @@ class Split < ActiveRecord::Base
 
   before_validation :cast_registry
 
+  scope :for_presentation, ->(app_build: nil) do
+    app_build.present? ? for_app_build(app_build) : active
+  end
+
+  scope :for_app_build, ->(app_build) do
+    active(as_of: app_build.built_at)
+      .with_feature_completeness(app_build)
+  end
+
   scope :active, ->(as_of: nil) do
     as_of ? where('splits.finished_at is null or splits.finished_at > ?', as_of) : where(finished_at: nil)
+  end
+
+  scope :with_feature_completeness, ->(app_build) do
+    select(column_names).readonly
+      .select(<<~SQL)
+        (
+          select splits.feature_gate = false
+          or exists (
+            select 1 from app_feature_completions fc
+            where fc.app_id = #{connection.quote(app_build.app_id)}
+            and fc.split_id = splits.id
+            and fc.version <= #{connection.quote(app_build.version.to_pg_array)}
+          )
+        ) as feature_complete
+      SQL
   end
 
   enum platform: %i(mobile desktop)
@@ -80,6 +104,20 @@ class Split < ActiveRecord::Base
 
   def create_decision!(params = {})
     build_decision(params).tap(&:save!)
+  end
+
+  def registry
+    if respond_to?(:feature_complete?) && !feature_complete?
+      super.each_with_object({}) do |(k, v), h|
+        if k == "false"
+          h[k] = 100
+        else
+          h[k] = 0
+        end
+      end
+    else
+      super
+    end
   end
 
   private
--- a/app/models/split.rb
+++ b/app/models/split.rb
@@ -132,7 +132,7 @@ class Split < ActiveRecord::Base
   def knock_out_weightings(registry_hash, to: "false")
     found_key = false
     knocked_out_registry = registry_hash.each_with_object({}) do |(k, _), h|
-      h[k] = (k.to_s == to ? (found_key = true && 100) : 0)
+      h[k] = (k.to_s == to.to_s ? (found_key = true && 100) : 0)
     end
     if found_key
       knocked_out_registry
--- a/app/models/split.rb
+++ b/app/models/split.rb
@@ -33,11 +33,14 @@ class Split < ActiveRecord::Base
   end
 
   scope :with_feature_completeness_for, ->(app_build) do
-    select(column_names)
+    previous_selects = all.arel.projections
+    except(:select)
       .select(
+        previous_selects,
         excluding_incomplete_features_for(app_build)
+          .except(:select)
           .select(1)
-          .from('(select null) as dual') # replace generated `FROM splits` which shadows us with a no-op "table"
+          .tap { |s| s.arel.ast.cores.first.from = nil } # inherit `splits` table from parent scope instead of doubling up
           .exists
           .as('feature_complete')
       )
--- a/app/models/split.rb
+++ b/app/models/split.rb
@@ -32,19 +32,25 @@ class Split < ActiveRecord::Base
     as_of ? where('splits.finished_at is null or splits.finished_at > ?', as_of) : where(finished_at: nil)
   end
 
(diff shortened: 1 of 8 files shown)


app/models/split.rb
```diff
@@ -17,13 +17,50 @@ class Split < ActiveRecord::Base
...
+  end
+
+  scope :excluding_incomplete_features_for, ->(app_build) do
+    where(arel_excluding_incomplete_features_for(app_build))
+  end
+
+  class << self
+    private
+
+    def arel_excluding_incomplete_features_for(app_build)
```
jmileham: The reationale for using an arel instead of a scope here is so it could be shared between the subselect above that needs to have no from clause in order to add a boolean pseudocolumn, and the more normal-looking filtering we do in `exclude_incomplete_features_for`.

A note on the kind of finicky language - calling these things `excluding_incomplete_features` instead of `for_complete_features` or something: The rationale is that there are two paths here - either it's a feature gate, and it's complete, or it's not a feature gate at all (not a feature). Anyway, lemme know if you hate that.

app/models/split.rb
```diff
@@ -82,8 +119,29 @@ def create_decision!(params = {})
...
+      super
+    end
+  end
+
   private
 
+  def knock_out_weightings(registry_hash, to: "false")
+    found_key = false
+    knocked_out_registry = registry_hash.each_with_object({}) do |(k, _), h|
+      h[k] = (k.to_s == to ? (found_key = true && 100) : 0)
```
jmileham:
```suggestion
      h[k] = (k.to_s == to.to_s ? (found_key = true && 100) : 0)
```

app/models/split.rb
```diff
@@ -17,13 +17,50 @@ class Split < ActiveRecord::Base
...
+    previous_selects = all.arel.projections
+    except(:select)
+      .select(
+        previous_selects,
+        Arel::SelectManager.new
+      .where(arel_excluding_incomplete_features_for(app_build).not)
+          .exists
+          .as('feature_incomplete')
+      )
+      .readonly
```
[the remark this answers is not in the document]
jmileham: It’s again got more to do with the modified behavior of the registry getter for instances fetched from this scope. If somebody took the knocked out registry for presentation to a specific app build and hard set the actual registry to match that registry, it’d be bad news bears.

app/models/split.rb
```diff
@@ -17,13 +17,50 @@ class Split < ActiveRecord::Base
   validate :registry_weights_must_be_integers
   validate :registry_must_have_winning_variant_if_decided
 
+  enum platform: %i(mobile desktop)
+
   before_validation :cast_registry
 
+  scope :for_presentation, ->(app_build: nil) do
```
smudge: Where will we actually accept the `app_build` parameter?

app/models/split.rb
```diff
@@ -82,8 +119,34 @@ def create_decision!(params = {})
     build_decision(params).tap(&:save!)
   end
 
+  def registry
+    if try(:feature_incomplete?)
```
smudge: Is this shorthand for `respond_to?(:feature_complete?) && feature_incomplete?`. Might be worth adding a comment explaining that this field is injected by one of the scopes.

app/models/split.rb
```diff
@@ -82,8 +119,34 @@ def create_decision!(params = {})
...
+    else
+      super
+    end
+  end
+
   private
 
+  def knock_out_weightings(registry_hash, to: "false")
+    target_variant = to.to_s
+    knock_out_weightings_if_possible(registry_hash, to: target_variant) || log_knockout_error(target_variant) && registry_hash
```
smudge: Nitpicking on the control flow here a little. It's very compact, but really what's happening is that this:

```
f(A) || log_error && A
```

Is doing some immediate-turnaround nil-handling for `f`, whose last line looks like

```
found_key ? A' : nil
```

When combined, those could look more like this:

```
if found_key
  A'
else
  log_error
  A
end
```

To me, that groks a bit more immediately, and also has the nice property of not caring about the return value of `log_error`.

smudge: A more rubyesque version of that last conditional could look more like this:

```ruby
f(A) || A
```

Where `f` looks like this:

```ruby
found_key? A' : log_error && nil
```

In my mind that does lose a little grokability again, but it might be a nice middle-ground between the existing, compact control flow, and a variation that doesn't care about whether log_error returns `true`

app/models/split.rb
```diff
@@ -82,8 +119,34 @@ def create_decision!(params = {})
...
+  end
+
+  def knock_out_weightings_if_possible(registry_hash, to:)
+    found_key = false
+    knocked_out_registry = registry_hash.each_with_object({}) do |(k, _), h|
+      h[k] = (k.to_s == to ? (found_key = true && 100) : 0)
```
smudge: If only `transform_values` let you peek at the key...

app/models/split.rb
```diff
@@ -82,8 +119,34 @@ def create_decision!(params = {})
...
+
   private
 
+  def knock_out_weightings(registry_hash, to: "false")
+    target_variant = to.to_s
+    knock_out_weightings_if_possible(registry_hash, to: target_variant) || log_knockout_error(target_variant) && registry_hash
+  end
+
+  def knock_out_weightings_if_possible(registry_hash, to:)
+    found_key = false
```
smudge: What if we expressed this as `found_key = registry_hash.key?(to)` and short-circuit out of the whole thing?

smudge: (is the concern that we need the `to_s` on all the keys? Can we `stringify_keys` first? Also, if they're coming straight from the DB, won't they always be strings?)

app/models/split.rb
jmileham: Probably right in the controller unless I build a decorator that takes the request bits and plumbs them in

app/models/split.rb
jmileham: Yeah. And yeah!

app/models/split.rb
jmileham: Yeah, I was threading a needle of rubocop line length here without adding a class, I had all those factoring along the way iirc. Note that your last suggestion would need a second arg to f, it’s probably an object envy issue if I really wanna make it work nice. Should I do that?

app/models/split.rb
jmileham: I get that this isn’t the prettiest on the inside. Lots of options, I think. I just suspect this code will never change after this. I’ve already got the function I need to knock out non-feature gates for remote kill scenarios.

jmileham: Re always be strings, the object has setters, I’m just shooting for the principle of least surprise if somebody makes a weird usage decision later. I can’t make them not set the value of registry, that’s not what readonly means iirc; just resists saves.

app/models/split.rb
```diff
@@ -82,8 +119,34 @@ def create_decision!(params = {})
...
+    if try(:feature_incomplete?)
+      knock_out_weightings(super)
+    else
+      super
+    end
+  end
+
   private
 
+  def knock_out_weightings(registry_hash, to: "false")
```
smudge: What is the plan for being able to go with something other than `'false'`

app/models/split.rb
smudge: So, I'm like 99% sure that `(found_key = true && 100)` will set `found_key` to `100`. Which, to be fair, still quacks the right way.

app/models/split.rb
```diff
@@ -82,8 +119,34 @@ def create_decision!(params = {})
     build_decision(params).tap(&:save!)
   end
 
+  def registry
+    if try(:feature_incomplete?)
+      knock_out_weightings(super)
```
smudge: `knock_out` => `mask`? (`conceal`? `override`?) In my first reading of this I sort of assumed "knock-out" had a more specific meaning, but I'm still not quite sure what that is -- I assume it's intended to be more "to make inoperative" and less "to make unconscious"?

app/models/split.rb
smudge: Feels _almost_ like `registry` wants to deserialize to a value object that can perform all of this knock-out magic internally. (Some kind of hash-like that accepts the additional context of `feature_incomplete` and whatever provides the `to` value.) But that said, I'm leaning more towards leaving it as-is (as you said, this code is not going to need to change much), unless something about that idea resonates with you.

app/models/split.rb
jmileham: Remote kills will support arbitrary splits because it’s basically an “oh god; this version range has a catastrophic bug in it - what can i grab onto to mitigate?” Tool, so you might e.g. go to 100% treatment, 100% control, 100% whatevs. But feature incomplete is for conventional feature gates with true/false variants.

app/models/split.rb
jmileham: Good eye. Yeah; I might just lean into that unless you have feels. The test coverage gets me comfortable with behavior and against regressions, but you tell me.

app/models/split.rb
jmileham: Mask to inoperative is what I’m hoping knock out means. It is only exposed in effectively private API. To consumers it’s part of the presentation contract and just works

app/models/split.rb
jmileham: Yeah that was what I was suggesting re object envy - extracting registry to a configurable-at-instantiatiation immutable hash-quackalike would be the nice version, but feels like wasted effort and global complexity in this context.

app/models/split.rb
smudge: If it were a more active codebase and I knew someone was going to have to come back through in the near term, I'd suggest expanding it into a full `if ... else` with an `h[k] =` assignment in each branch. As-is, it sort of feels like you're tucking away all of this knock-out logic into as small a corner of this class as it can fit. But as you said, as long as there are tests, maybe it's fine to let it be and see how far it can take us.

app/models/split.rb
```diff
@@ -82,8 +119,34 @@ def create_decision!(params = {})
...
+    found_key = false
+    knocked_out_registry = registry_hash.each_with_object({}) do |(k, _), h|
+      h[k] = (k.to_s == to ? (found_key = true && 100) : 0)
+    end
+    found_key ? knocked_out_registry : nil
+  end
+
+  def log_knockout_error(to)
+    logger.error "Failed to knock out weightings of split #{name.inspect} because variant #{to.inspect} not found."
+    true
```
smudge: Anything that inherits from the stdlib `Logger` should already return `true` here, but I suppose there's always the odd chance that someone's custom logger causes bizarre behavior in the spot where we chain it onto an `&&`.

jmileham: yup, built for safety.

spec/models/app_feature_completion_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      expect { described_class.satisfied_by(app_build).to_a }.to raise_error(/missing FROM-clause entry for table "splits"/)
+    end
+
+    it "returns a record for an app_build with the same app_id and version" do
+      fc = FactoryBot.create(:app_feature_completion, app: app, split: split, version: "1.0")
+
+      expect(described_class.joins(:split).satisfied_by(app_build)).to include(fc)
+    end
+
+    it "doesn't return a record for an app_build with a lower version" do
```
smudge: Worth adding the "higher version" case too, to make sure the query doesn't get changed from `lteq` to `eq`?

spec/models/split_spec.rb
```diff
@@ -208,4 +208,162 @@
       expect(described_class.active(as_of: Time.zone.now)).not_to include(split)
     end
   end
+
+  describe ".with_feature_incomplete_knockouts_for" do
+    let(:app) { FactoryBot.create(:app) }
+    let(:app_build) { app.define_build(built_at: Time.zone.now, version: "1.0") }
+
+    it "respects existing selects" do
```
smudge: 😎

spec/models/split_spec.rb
```diff
@@ -208,4 +208,162 @@
...
+      split = FactoryBot.create(:split, feature_gate: true)
+      expect(Split.with_feature_incomplete_knockouts_for(app_build).find(split.id)).to be_feature_incomplete
+    end
+
+    it "is backed by AppFeatureCompletion.satisfied_by" do
+      allow(AppFeatureCompletion).to receive(:satisfied_by).and_call_original
+
+      Split.with_feature_incomplete_knockouts_for(app_build)
+
+      expect(AppFeatureCompletion).to have_received(:satisfied_by).with(app_build)
```
smudge: This doesn't _quite_ prove that it was effectively merged into the query, just that the code was executed. But aside from fully testing the integration the only other thing here would be to inspect the SQL or something.

smudge: (similar arguments would apply elsewhere too, not sure how far we would want to take that.)

smudge:
<< domain LGTM && platform TAFN

Just a couple questions about the tests. Mainly I want to make sure if we're leaning in on them for assurances about all of these internal API nuances, that we're confident that the areas where we're stubbing or expecting method calls (in scopes) aren't opening us to logical regressions in the fully-formed queries. (As in, just because a collaborator gets called doesn't mean it gets properly incorporated into the query). I'm not suggesting a particular change to the way these tests are written, just curious to hear your thoughts.

spec/models/split_spec.rb
jmileham: Yeah, for sure not proof. The testing strategy generally for composed query logic is attempting to satisfy occam's razor that "if it works like this and calls that thing that is purpose built to create that behavior, it probably works like this because it is using that thing" - if you don't feel that we're comfortably hitting that balance, I think that'd be where the heat would be.

spec/models/app_feature_completion_spec.rb
jmileham: yup!

app/models/split.rb
jmileham: rubocop will cut you if the last line of each fork assigns to the same var.

app/models/split.rb
jmileham: (note that you can use a remote kill instead if you accidentally screwed up the variant names on a feature gate - it's just a more powerful version of the tool that requires more configuration because it needs a first version and identity (a reason slug) for future mutability)

jmileham:
bump!

spec/models/split_spec.rb
smudge: There may be a way that I just haven't discovered of cracking open an AR relation and comparing its innards with something else, which would enable something like this:

```ruby
query = Split.with_feature_incomplete_knockouts_for(app_build)
query_part = AppFeatureCompletion.satisifed_by(app_build)
expect(query).to have_incorporated(query_part)
```

I've also controversially done something like this in the past:

```ruby
expect(query.to_sql).to include(<actual subquery SQL>)
```

But I'm not proposing that we do that here. I'd say for now I'm comfortable with the balance we're hitting, even if it's not an ideal state.

smudge:
platform LGTM

I've bookmarked this PR and titled it "Practical Arel Reference"

spec/models/split_spec.rb
jmileham: Yeah, I think that's a good strategy too, and IIRC I've done it before also. It'd probably be a bit more brittle (irrelevant if this is low-rate-of-change), but stronger confidence for sure. I'm happy with the confidence level as-is though, so yeah, let's rock it. Thanks for the deep dive.

PR #104, opened 2019-04-09
Hard assignments

jmileham:
### Summary

Adds a force flag to assignments which will later be plumbed into assignment hiding logic for app_feature_incompletions and app_remote_kills. Using this column for presentation comes later, and that's fine because no endpoints currently available will be impacted by the force flag.

### Other Information

The tests here are very high level, but actually give a lot of confidence. There were still-other tests that failed along the way and showed me where I was under-implementing this feature. I'm not sure the test strategy is super opinionated for all these models in general, but this testing strategy feels coherent with the way tests had been written before. What are your thoughts on that?

There's also a detail here that's kind of annoying - I want to add a not-null column to a potentially-very-large table (assignments). The safest way to do that is not supported by ActiveRecord migrations. I'm proposing that I do it the safe way, and then lie about how I did it in the schema file, intentionally introducing a dissonance between how the constraint is implemented between the production database (of an existing install that is migrated up) and any database built from the latest schema.rb.

I'll have comments below about this.

/domain @alederman @bslakter for migration questions

### DIFF
--- a/app/models/arbitrary_assignment_creation.rb
+++ b/app/models/arbitrary_assignment_creation.rb
@@ -1,4 +1,6 @@
 class ArbitraryAssignmentCreation
+  attr_reader :visitor_id, :split_name, :variant, :bulk_assignment_id, :context
+
   def initialize(params = {})
     @visitor_id = params[:visitor_id]
     @split_name = params[:split_name]
@@ -6,6 +8,7 @@ class ArbitraryAssignmentCreation
     @mixpanel_result = params[:mixpanel_result]
     @bulk_assignment_id = params[:bulk_assignment_id]
     @context = params[:context]
+    @force = params[:force] if params.key?(:force)
   end
 
   def save!
@@ -18,6 +21,10 @@ class ArbitraryAssignmentCreation
 
   def self.create!(params)
     new(params).tap(&:save!)
+  end
+
+  def force
+    instance_variable_defined?(:@force) ? @force : false
   end
 
   private
@@ -78,7 +85,8 @@ class ArbitraryAssignmentCreation
       variant: variant,
       mixpanel_result: mixpanel_result,
       bulk_assignment_id: bulk_assignment_id_for_save,
-      context: context_for_save
+      context: context_for_save,
+      force: force
     }
   end
 
@@ -110,6 +118,4 @@ class ArbitraryAssignmentCreation
       context_for_save
     end
   end
-
-  attr_reader :visitor_id, :split_name, :variant, :bulk_assignment_id, :context
 end
--- a/app/models/arbitrary_assignment_creation.rb
+++ b/app/models/arbitrary_assignment_creation.rb
@@ -1,15 +1,25 @@
 class ArbitraryAssignmentCreation
-  attr_reader :visitor_id, :split_name, :variant, :bulk_assignment_id, :context
+  attr_reader :visitor_id, :split_name, :variant, :bulk_assignment_id, :context, :force
 
-  def initialize(params = {})
-    @visitor_id = params[:visitor_id]
-    @split_name = params[:split_name]
-    @variant = params[:variant]
-    @mixpanel_result = params[:mixpanel_result]
-    @bulk_assignment_id = params[:bulk_assignment_id]
-    @context = params[:context]
-    @force = params[:force] if params.key?(:force)
+  # rubocop:disable Metrics/ParameterLists
+  def initialize(
+    visitor_id: nil,
+    split_name: nil,
+    variant: nil,
+    mixpanel_result: nil,
+    bulk_assignment_id: nil,
+    context: nil,
+    force: false
+  )
+    @visitor_id = visitor_id
+    @split_name = split_name
+    @variant = variant
+    @mixpanel_result = mixpanel_result
+    @bulk_assignment_id = bulk_assignment_id
+    @context = context
+    @force = force
   end
+  # rubocop:enable Metrics/ParameterLists
 
   def save!
     if superseding?
@@ -20,11 +30,7 @@ class ArbitraryAssignmentCreation
   end
 
   def self.create!(params)
-    new(params).tap(&:save!)
-  end
-
-  def force
-    instance_variable_defined?(:@force) ? @force : false
+    new(params.to_h.symbolize_keys).tap(&:save!)
   end
 
   private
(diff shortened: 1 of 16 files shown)


db/migrate/20190408145933_add_not_null_constraints.rb
```diff
@@ -0,0 +1,13 @@
+class AddNotNullConstraints < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+  def up
+    execute "ALTER TABLE assignments ADD CONSTRAINT assignments_force_check CHECK (force IS NOT NULL) NOT VALID;"
+    execute "ALTER TABLE previous_assignments ADD CONSTRAINT previous_assignments_force_check CHECK (force IS NOT NULL) NOT VALID;"
+    execute "ALTER TABLE bulk_assignments ADD CONSTRAINT bulk_assignments_force_check CHECK (force IS NOT NULL) NOT VALID;"
```
jmileham: These constraints will not be represented in the schema.rb because they're check constraints, not usual column not null declarations. That's sort of the crux of why I'm doing weird stuff here.

db/migrate/20190409164425_validate_not_null_constriants.rb
```diff
@@ -0,0 +1,12 @@
+class ValidateNotNullConstriants < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+  def up
+    execute "ALTER TABLE assignments VALIDATE CONSTRAINT assignments_force_check;"
+    execute "ALTER TABLE previous_assignments VALIDATE CONSTRAINT previous_assignments_force_check;"
+    execute "ALTER TABLE bulk_assignments VALIDATE CONSTRAINT bulk_assignments_force_check;"
```
jmileham: Here's the validate, which takes a long time, but also only out a very ungrabby lock - only conflicts with DDL on the table.

db/schema.rb
```diff
@@ -66,6 +66,7 @@
     t.uuid "bulk_assignment_id"
     t.uuid "visitor_supersession_id"
     t.string "context"
+    t.boolean "force", default: false, null: false
```
jmileham: These `null: false` values are hand-edited, so that future users will get the good version, and current users will get the safe version.

We'll have to be careful that subsequent PRs from people who had an existing local DB that did the migration version do not accidentally remote this constraint.

jmileham:
Bulk assignment form looks like this now:

Screen Shot 2019-04-09 at 10 34 53 AM

db/migrate/20190408145933_add_not_null_constraints.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+  def down
+    execute "ALTER TABLE assignments ADD CONSTRAINT assignments_force_check;"
```
alederman: These should be `DROP`s.

jmileham: good eye!

alederman:
<< domainLGTM for schema related changes.

app/models/arbitrary_assignment_creation.rb
```diff
@@ -1,11 +1,14 @@
...
+  attr_reader :visitor_id, :split_name, :variant, :bulk_assignment_id, :context
+
   def initialize(params = {})
     @visitor_id = params[:visitor_id]
     @split_name = params[:split_name]
     @variant = params[:variant]
     @mixpanel_result = params[:mixpanel_result]
     @bulk_assignment_id = params[:bulk_assignment_id]
     @context = params[:context]
+    @force = params[:force] if params.key?(:force)
```
samandmoore: why the condition here?

app/models/arbitrary_assignment_creation.rb
```diff
@@ -20,6 +23,10 @@ def self.create!(params)
     new(params).tap(&:save!)
   end
 
+  def force
+    instance_variable_defined?(:@force) ? @force : false
```
samandmoore: if you removed the condition above then this could be `@force == true`, couldn't it?

app/models/bulk_reassignment.rb
```diff
@@ -48,22 +49,24 @@ def insert_previous_assignments
...
   end
 
-  def update_assignments
+  def update_assignments # rubocop:disable Metrics/AbcSize
     connection.execute(<<-SQL)
       UPDATE assignments SET
         variant = #{connection.quote(bulk_assignment.variant)},
         updated_at = #{connection.quote(now)},
         mixpanel_result = NULL,
         bulk_assignment_id = #{connection.quote(bulk_assignment)},
```
samandmoore: not from this diff, but was this supposed to be `bulk_assignment.id`?

app/models/bulk_assignment_creation.rb
```diff
@@ -83,7 +93,10 @@ def identifier_type
 
   def existing_assignments
     ensure_identifiers
-    Assignment.where(visitor_id: existing_visitor_ids, split: split).where.not(variant: variant)
+    Assignment.where(visitor_id: existing_visitor_ids, split: split).where.not(
+      Assignment.arel_table[:variant].eq(variant)
+      .and(Assignment.arel_table[:force].eq(force))
+    )
```
samandmoore: this change is the only thing i can see in here that doesn't appear to be tested. worth adding a test that a preexisting "force" assignment doesn't change after a bulk assignment? (i think that's what this is intending)

samandmoore:
<<domain platform

just a few questions about that force ivar and maybe one test case we should add. 

domainlgtm platformlgtm

app/models/arbitrary_assignment_creation.rb
jmileham: Because I'm distinguishing between not-passed and expressly-set-to-null so I can default to false.

app/models/bulk_assignment_creation.rb
samandmoore: oh no. i see that this is saying that if the inputs are "variant: yes, force: true" exclude existing assignments that match that exactly. so, if there's an existing assignment for "variant: yes, force: false" then this will still scoop that one up. coolcoolcool.

app/models/arbitrary_assignment_creation.rb
jmileham: I'm interested in validating that this is not passed as null

app/models/bulk_reassignment.rb
jmileham: it's the behavior of quoting an activerecord and IIRC I chose it explicitly, but it does seem a little fancy.

samandmoore: yea, i suspected that, but it just seemed unnecessary to me since you can just ask for the id and it's less surprising to me that way. but i don't care that much.

PR #105, opened 2019-04-09
Force beats feature incomplete

jmileham:
### Summary

~~Submitting this PR before previous merges. Much less controversial, but the body of it is only [the latest commit](https://github.com/Betterment/test_track/pull/105/commits/a8eef2280699120b650e67dcc603738035f14851) - the rest is a duplicate of #104.~~

### DIFF
--- a/app/models/split.rb
+++ b/app/models/split.rb
@@ -47,19 +47,11 @@ class Split < ActiveRecord::Base
       .readonly
   end
 
-  scope :excluding_incomplete_features_for, ->(app_build) do
-    where(arel_excluding_incomplete_features_for(app_build))
-  end
-
-  class << self
-    private
-
-    def arel_excluding_incomplete_features_for(app_build)
-      Arel::Nodes::Or.new(
-        arel_table[:feature_gate].eq(false),
-        AppFeatureCompletion.select(1).satisfied_by(app_build).arel.exists
-      )
-    end
+  def self.arel_excluding_incomplete_features_for(app_build)
+    Arel::Nodes::Or.new(
+      arel_table[:feature_gate].eq(false),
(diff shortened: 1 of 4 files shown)


samandmoore:
<<domain platform

app/models/split.rb
```diff
@@ -47,19 +47,11 @@ class Split < ActiveRecord::Base
...
-  class << self
-    private
-
-    def arel_excluding_incomplete_features_for(app_build)
-      Arel::Nodes::Or.new(
-        arel_table[:feature_gate].eq(false),
-        AppFeatureCompletion.select(1).satisfied_by(app_build).arel.exists
-      )
-    end
+  def self.arel_excluding_incomplete_features_for(app_build)
```
samandmoore: @smudge and i were thinking that it might be helpful if for each of these scopes to come with an example SQL fragment as a comment above the scope, thoughts?

samandmoore:
domainlgtm platformlgtm

app/models/split.rb
jmileham: I'm a little squeamish about dropping a comment with the SQL it generates today because nobody will keep it correct, but here's the result of calling this:

```ruby
puts Split.arel_excluding_incomplete_features_for(App.first.define_build(version: "1.0", built_at: Time.zone.now)).to_sql
```

> "splits"."feature_gate" = 'f' OR EXISTS (SELECT 1 FROM "app_feature_completions" WHERE ("app_feature_completions"."split_id" = "splits"."id" AND "app_feature_completions"."app_id" = '0302560e-754f-4cdd-bdff-35c543650def' AND "app_feature_completions"."version" <= '{1,0}'))

PR #106, opened 2019-04-11
Remote kill

jmileham:
### Summary

Adds app_remote_kill functionality. Will annotate below.

### DIFF
--- a/app/models/app_feature_completion.rb
+++ b/app/models/app_feature_completion.rb
@@ -7,12 +7,9 @@ class AppFeatureCompletion < ActiveRecord::Base
   validates :app, :split, :version, presence: true
   validates :split, uniqueness: { scope: :app }
 
-  # This scope requires you to BYO `splits` FROM clause either via join or an
-  # outer scope, if using this as a subselect.
   scope :satisfied_by, ->(app_build) do
     where(
-      arel_table[:split_id].eq(Split.arel_table[:id])
-      .and(arel_table[:app_id].eq(app_build.app_id))
+      arel_table[:app_id].eq(app_build.app_id)
       .and(arel_table[:version].lteq(app_build.version))
     )
   end
--- a/app/models/app_remote_kill.rb
+++ b/app/models/app_remote_kill.rb
@@ -1,0 +1,71 @@
+class AppRemoteKill < ActiveRecord::Base
+  belongs_to :app
+  belongs_to :split
+
+  attribute :first_bad_version, :app_version
+  attribute :fixed_version, :app_version
+
+  validates :app, :split, :reason, :override_to, :first_bad_version, presence: true
+  validates :reason, uniqueness: { scope: %i(app split) }
+
+  validate :override_to_must_exist
+  validate :fixed_version_must_be_greater_than_first_bad_version
+  validate :must_not_overlap_existing
+
+  # This scope requires you to BYO `splits` FROM clause either via join or an
+  # outer scope, if using this as a subselect.
+  scope :affecting, ->(app_build) do
+    where(
+      arel_table[:split_id].eq(Split.arel_table[:id])
+      .and(arel_table[:app_id].eq(app_build.app_id))
+      .and(arel_table[:first_bad_version].lteq(app_build.version))
+      .and(
+        Arel::Nodes::Or.new(
+          arel_table[:fixed_version].eq(nil),
+          arel_table[:fixed_version].gt(app_build.version)
+        )
+      )
+    )
+  end
+
+  scope :overlapping, ->(other) do
+    where(app_id: other.app_id, split_id: other.split_id)
+      .where.not(id: other.id)
+      .where(
+        if other.fixed_version.nil?
+          Arel::Nodes::True.new
+        else
+          arel_table[:first_bad_version].lt(other.fixed_version)
+        end
+        .and(
+          Arel::Nodes::Grouping.new(
+            Arel::Nodes::Or.new(
+              arel_table[:fixed_version].eq(nil),
+              arel_table[:fixed_version].gt(other.first_bad_version)
+            )
+          )
+        )
+      )
+  end
+
+  def override_to_must_exist
+    return unless split
+
+    errors.add(:override_to, "must exist in split's current variants") unless split.has_variant?(override_to)
+  end
+
+  def fixed_version_must_be_greater_than_first_bad_version
+    return if fixed_version.nil?
+
+    errors.add(:fixed_version, "must be greater than first bad version") unless fixed_version > first_bad_version
+  end
+
+  def must_not_overlap_existing
+    return if first_bad_version.nil? || split.nil? || app.nil?
+
+    overlapping = self.class.overlapping(self)
+    return if overlapping.empty?
+
+    errors.add(:base, "must not overlap existing app remote kills: #{overlapping.map(&:reason).join(', ')}")
+  end
+end
--- a/app/models/app_remote_kill.rb
+++ b/app/models/app_remote_kill.rb
@@ -12,19 +12,11 @@ class AppRemoteKill < ActiveRecord::Base
   validate :fixed_version_must_be_greater_than_first_bad_version
   validate :must_not_overlap_existing
 
-  # This scope requires you to BYO `splits` FROM clause either via join or an
-  # outer scope, if using this as a subselect.
   scope :affecting, ->(app_build) do
     where(
-      arel_table[:split_id].eq(Split.arel_table[:id])
-      .and(arel_table[:app_id].eq(app_build.app_id))
+      arel_table[:app_id].eq(app_build.app_id)
       .and(arel_table[:first_bad_version].lteq(app_build.version))
-      .and(
-        Arel::Nodes::Or.new(
-          arel_table[:fixed_version].eq(nil),
-          arel_table[:fixed_version].gt(app_build.version)
-        )
-      )
+      .and(arel_fixed_version_is_null_or_greater_than(app_build.version))
     )
   end
 
@@ -37,15 +29,17 @@ class AppRemoteKill < ActiveRecord::Base
         else
           arel_table[:first_bad_version].lt(other.fixed_version)
         end
-        .and(
-          Arel::Nodes::Grouping.new(
-            Arel::Nodes::Or.new(
-              arel_table[:fixed_version].eq(nil),
-              arel_table[:fixed_version].gt(other.first_bad_version)
-            )
-          )
-        )
+        .and(arel_fixed_version_is_null_or_greater_than(other.first_bad_version))
       )
+  end
+
+  def self.arel_fixed_version_is_null_or_greater_than(version)
+    Arel::Nodes::Grouping.new(
+      Arel::Nodes::Or.new(
+        arel_table[:fixed_version].eq(nil),
+        arel_table[:fixed_version].gt(version)
+      )
+    )
   end
 
   def override_to_must_exist
--- a/app/models/app_remote_kill.rb
+++ b/app/models/app_remote_kill.rb
@@ -12,12 +12,26 @@ class AppRemoteKill < ActiveRecord::Base
   validate :fixed_version_must_be_greater_than_first_bad_version
   validate :must_not_overlap_existing
 
-  scope :affecting, ->(app_build) do
+  scope :affecting, ->(app_build, override: false, overridden_at: nil) do
     where(
       arel_table[:app_id].eq(app_build.app_id)
       .and(arel_table[:first_bad_version].lteq(app_build.version))
       .and(arel_fixed_version_is_null_or_greater_than(app_build.version))
+      .and(
+        Arel::Nodes::Grouping.new(
+          arel_is_false(override)
+          .or(arel_table[:updated_at].gt(overridden_at))
+        )
+      )
     )
+  end
+
+  class << self
+    private
+
+    def arel_is_false(value_or_arel)
+      Arel::Nodes::Grouping.new(Arel::Nodes::False.new).eq(value_or_arel)
+    end
   end
 
   scope :overlapping, ->(other) do
@@ -35,10 +49,8 @@ class AppRemoteKill < ActiveRecord::Base
 
   def self.arel_fixed_version_is_null_or_greater_than(version)
     Arel::Nodes::Grouping.new(
-      Arel::Nodes::Or.new(
-        arel_table[:fixed_version].eq(nil),
-        arel_table[:fixed_version].gt(version)
-      )
+      arel_table[:fixed_version].eq(nil)
+      .or(arel_table[:fixed_version].gt(version))
     )
   end
 
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -48,7 +48,14 @@ class Assignment < ActiveRecord::Base
   end
 
   scope :excluding_remote_kills_for, ->(app_build) do
-    joins(:split).merge(Split.excluding_remote_kills_for(app_build))
+    joins(:split).where(
+      Arel::Nodes::Grouping.new(
+        Arel::Nodes::Or.new(
+          arel_table[:force].eq(true),
+          Split.arel_excluding_remote_kills_for(app_build)
+        )
+      )
+    )
   end
 
   def variant_detail
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -37,9 +37,11 @@ class Assignment < ActiveRecord::Base
 
   scope :excluding_incomplete_features_for, ->(app_build) do
     joins(:split).where(
-      Arel::Nodes::Or.new(
-        arel_table[:force].eq(true),
-        Split.arel_excluding_incomplete_features_for(app_build)
+      Arel::Nodes::Grouping.new(
+        Arel::Nodes::Or.new(
+          arel_table[:force].eq(true),
+          Split.arel_excluding_incomplete_features_for(app_build)
+        )
       )
     )
   end
@@ -81,6 +83,6 @@ class Assignment < ActiveRecord::Base
   def variant_must_exist
     return unless split
 
-    errors.add(:variant, "must be specified in split's current variations") unless split.has_variant?(variant)
+    errors.add(:variant, "must be specified in split's current variants") unless split.has_variant?(variant)
   end
 end
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -25,6 +25,7 @@ class Assignment < ActiveRecord::Base
   scope :for_app_build, ->(app_build) do
     for_active_splits(as_of: app_build.built_at)
       .excluding_incomplete_features_for(app_build)
+      .excluding_remote_kills_for(app_build)
   end
 
   scope :excluding_decision_overrides, -> do
@@ -44,6 +45,10 @@ class Assignment < ActiveRecord::Base
         )
       )
     )
+  end
+
+  scope :excluding_remote_kills_for, ->(app_build) do
+    joins(:split).merge(Split.excluding_remote_kills_for(app_build))
   end
 
   def variant_detail
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -39,22 +39,15 @@ class Assignment < ActiveRecord::Base
   scope :excluding_incomplete_features_for, ->(app_build) do
     joins(:split).where(
       Arel::Nodes::Grouping.new(
-        Arel::Nodes::Or.new(
-          arel_table[:force].eq(true),
-          Split.arel_excluding_incomplete_features_for(app_build)
-        )
+        arel_table[:force].eq(true)
+        .or(Split.arel_excluding_incomplete_features_for(app_build))
       )
     )
   end
 
   scope :excluding_remote_kills_for, ->(app_build) do
     joins(:split).where(
-      Arel::Nodes::Grouping.new(
-        Arel::Nodes::Or.new(
-          arel_table[:force].eq(true),
-          Split.arel_excluding_remote_kills_for(app_build)
-        )
-      )
+      Split.arel_excluding_remote_kills_for(app_build, override: arel_table[:force], overridden_at: arel_table[:updated_at])
     )
   end
 
--- a/app/models/split.rb
+++ b/app/models/split.rb
@@ -28,6 +28,7 @@ class Split < ActiveRecord::Base
   scope :for_app_build, ->(app_build) do
     active(as_of: app_build.built_at)
       .with_feature_incomplete_knockouts_for(app_build)
+      .with_remote_kill_knockouts_for(app_build)
   end
 
   scope :active, ->(as_of: nil) do
@@ -56,6 +57,18 @@ class Split < ActiveRecord::Base
         .exists
       )
     )
+  end
+
+  scope :with_remote_kill_knockouts_for, ->(app_build) do
+    previous_selects = all.arel.projections
+    except(:select)
+      .select(
+        previous_selects,
+        AppRemoteKill.select(:override_to).affecting(app_build).arel
+          .where(AppRemoteKill.arel_table[:split_id].eq(arel_table[:id]))
+          .as('remote_kill_override_to')
+      )
+      .readonly
   end
 
   scope :excluding_remote_kills_for, ->(app_build) do
@@ -124,7 +137,9 @@ class Split < ActiveRecord::Base
   end
 
   def registry
-    if try(:feature_incomplete?) # This is a virtual attribute provided by the with_feature_incomplete_knockouts_for scope
+    if try(:remote_kill_override_to) # These are virtual attributes provided by the for_app_build scope
+      knock_out_weightings(super, to: remote_kill_override_to)
+    elsif try(:feature_incomplete?) # These are virtual attributes provided by the for_app_build scope
       knock_out_weightings(super)
     else
       super
--- a/app/models/split.rb
+++ b/app/models/split.rb
@@ -48,9 +48,21 @@ class Split < ActiveRecord::Base
   end
 
   def self.arel_excluding_incomplete_features_for(app_build)
-    Arel::Nodes::Or.new(
-      arel_table[:feature_gate].eq(false),
-      AppFeatureCompletion.select(1).satisfied_by(app_build).arel.exists
+    Arel::Nodes::Grouping.new(
+      Arel::Nodes::Or.new(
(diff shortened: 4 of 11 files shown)


jmileham:
Here's the actual diff of this PR: https://github.com/Betterment/test_track/pull/106/files/5be842c03dfb3b90111713ba800ee431295f81fc..8e748da119b94b1093e06cea303cc27eba8519f4

app/models/app_feature_completion.rb
```diff
@@ -7,12 +7,9 @@ class AppFeatureCompletion < ActiveRecord::Base
   validates :app, :split, :version, presence: true
   validates :split, uniqueness: { scope: :app }
 
-  # This scope requires you to BYO `splits` FROM clause either via join or an
-  # outer scope, if using this as a subselect.
```
jmileham: I realized this was a factoring problem - I can inject the dependency on splits at the point of use when merging into a scope on Split and make this a regular scope. Followed the same pattern on AppRemoteKill scope.

app/models/app_remote_kill.rb
```diff
@@ -0,0 +1,65 @@
+class AppRemoteKill < ActiveRecord::Base
+  belongs_to :app
+  belongs_to :split
+
+  attribute :first_bad_version, :app_version
+  attribute :fixed_version, :app_version
```
jmileham: In the design doc, this was gonna be `last_bad_version` but I realized that `fixed_version` gives it semantic parity with a feature completion's `version` - when you ship the bug fix, you can drop in your own version number without looking up the last version number that was bad. Also, this fails safe in a multi-branch development strategy where 0.9 was the last bad version, 1.0 contains the fix, and then 0.9.1 comes out to address another bug and it still has the original bug and wants the remote kill to remain active (it's certainly possible to edit the fixed_version if you want to say that 0.9.1 is fixed too, but this will default to not enabling the feature until the version you declared safe).

app/models/app_remote_kill.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  validates :app, :split, :reason, :override_to, :first_bad_version, presence: true
+  validates :reason, uniqueness: { scope: %i(app split) }
+
+  validate :override_to_must_exist
+  validate :fixed_version_must_be_greater_than_first_bad_version
+  validate :must_not_overlap_existing
+
+  scope :affecting, ->(app_build) do
```
jmileham: Where the feature incompletion scope was selecting feature completions satisfied by that app version (i.e. splits not knocked out/assignments not overridden for that version), here the scope selects the remote kills that *do* impact the app. This is because feature completion knockouts operate simply on the boolean absence or presence of satisfying feature completion (and always use `false` as the override variant for knockouts), whereas for remote kills, we need to join in the `override_to` column to know what variant to select (because remote kills can be used on any kind of split).

app/models/app_remote_kill.rb
```diff
@@ -0,0 +1,65 @@
...
+      .and(arel_fixed_version_is_null_or_greater_than(app_build.version))
+    )
+  end
+
+  scope :overlapping, ->(other) do
+    where(app_id: other.app_id, split_id: other.split_id)
+      .where.not(id: other.id)
+      .where(
+        if other.fixed_version.nil?
+          Arel::Nodes::True.new
```
jmileham: This is mostly to make the chaining easier - really I just want to only do the second half of this where clause if other doesn't have a fixed version. I don't think I'm costing PG much to process `TRUE AND [...]` vs `[...]`

app/models/app_remote_kill.rb
```diff
@@ -0,0 +1,65 @@
...
+        else
+          arel_table[:first_bad_version].lt(other.fixed_version)
+        end
+        .and(arel_fixed_version_is_null_or_greater_than(other.first_bad_version))
+      )
+  end
+
+  def self.arel_fixed_version_is_null_or_greater_than(version)
+    Arel::Nodes::Grouping.new(
+      Arel::Nodes::Or.new(
```
jmileham: TIL that Or nodes are just that - OR'd together nodes in the AST - they don't confer parens around the OR, which is very unsafe, but hey, that's how SQL works. So you'll notice that I wrapped Grouping nodes (Arel for "parens") around all the Or nodes I'd been building elsewhere.

app/models/app_remote_kill.rb
```diff
@@ -0,0 +1,65 @@
...
+  def override_to_must_exist
+    return unless split
+
+    errors.add(:override_to, "must exist in split's current variants") unless split.has_variant?(override_to)
+  end
+
+  def fixed_version_must_be_greater_than_first_bad_version
+    return if fixed_version.nil?
+
+    errors.add(:fixed_version, "must be greater than first bad version") unless fixed_version > first_bad_version
```
jmileham: Because fixed version is the first version *not* affected by this remote kill, `>` is what we want here, not `>=`, which would be nonsensical - declaring the same version bad and good at the same time.

app/models/assignment.rb
```diff
@@ -37,13 +38,19 @@ class Assignment < ActiveRecord::Base
...
   scope :excluding_incomplete_features_for, ->(app_build) do
     joins(:split).where(
-      Arel::Nodes::Or.new(
-        arel_table[:force].eq(true),
-        Split.arel_excluding_incomplete_features_for(app_build)
+      Arel::Nodes::Grouping.new(
+        Arel::Nodes::Or.new(
+          arel_table[:force].eq(true),
+          Split.arel_excluding_incomplete_features_for(app_build)
+        )
```
jmileham: Just adding a parens here for safety.

app/models/assignment.rb
```diff
@@ -37,13 +38,19 @@ class Assignment < ActiveRecord::Base
...
       )
     )
   end
 
+  scope :excluding_remote_kills_for, ->(app_build) do
+    joins(:split).merge(Split.excluding_remote_kills_for(app_build))
+  end
```
jmileham: This feels very normal for how fancy the generated SQL is.

app/models/assignment.rb
```diff
@@ -81,6 +88,6 @@ def previous_assignment_params
   def variant_must_exist
     return unless split
 
-    errors.add(:variant, "must be specified in split's current variations") unless split.has_variant?(variant)
+    errors.add(:variant, "must be specified in split's current variants") unless split.has_variant?(variant)
```
jmileham: Just sniping a naming inconsistency

app/models/split.rb
```diff
@@ -48,9 +49,33 @@ class Split < ActiveRecord::Base
...
-      arel_table[:feature_gate].eq(false),
-      AppFeatureCompletion.select(1).satisfied_by(app_build).arel.exists
+    Arel::Nodes::Grouping.new(
+      Arel::Nodes::Or.new(
+        arel_table[:feature_gate].eq(false),
+        AppFeatureCompletion.select(1).satisfied_by(app_build).arel
+        .where(AppFeatureCompletion.arel_table[:split_id].eq(arel_table[:id]))
+        .exists
+      )
+    )
```
jmileham: Just parens again

spec/models/app_feature_completion_spec.rb
```diff
@@ -2,39 +2,32 @@
...
-    let(:split) { FactoryBot.create(:split) }
     let(:app) { FactoryBot.create(:app) }
     let(:app_build) { app.define_build(built_at: Time.zone.now, version: "1.0") }
     let(:different_app) { FactoryBot.create(:app) }
 
-    it "blows up if you don't BYO splits table" do
-      FactoryBot.create(:app_feature_completion, app: app, split: split, version: "1.0")
-
-      expect { described_class.satisfied_by(app_build).to_a }.to raise_error(/missing FROM-clause entry for table "splits"/)
-    end
```
jmileham: Note to self: don't test implementation details :) That said, I was pretty sure this was gonna have to work like this forever when I wrote the test and wanted to make sure I didn't accidentally regress the split scoping or else it would've returned incorrect SQL all over the place. Instead now, we test the split join behavior in split_spec.

spec/models/split_spec.rb
```diff
@@ -282,45 +282,173 @@
...
+        fc = FactoryBot.create(:app_feature_completion, version: "1.1", split: subject)
+        app_build = fc.app.define_build(built_at: Time.zone.now, version: "1.1")
+
+        subject_with_knockouts = Split.for_app_build(app_build).find(subject.id)
+
+        expect(subject_with_knockouts).not_to be_feature_incomplete
+        expect(subject_with_knockouts.registry).to eq("true" => 50, "false" => 50)
+      end
+
+      it "prefers a remote kill over a feature incompletion if both are present" do
```
jmileham: Note that in the design doc, I was thinking that remote kills would conflict with feature incompletions, but I'm now realizing that you might want to override to a specific variant over a not-yet-released feature. For example, a feature flag with nonstandard variant names you realize too late. So now we allow them not to conflict and give remote kills precedence over feature completions. remote kills can't overlap one another, though, because it'd be impossible to tell which variant to override to without some annoying unsemantic tiebreaker logic.

app/models/app_remote_kill.rb
smudge: Feels like Arel is in need of a more concise DSL that also doesn't hide all of these features like AR's query builder.

smudge: (I know there have been a few attempts over the years, but even some aliasing to more terse method names might help.)

app/models/assignment.rb
jmileham: Oh this should respect force: true also. Nothing should prevent a determined tester from seeing a broken variant.

app/models/app_remote_kill.rb
jmileham: I just switched back to infix or syntax actually, which partially solves. I still need to use gnarly `Arel::Nodes::Grouping.new` but can just use `.or(other)` on a variety of arel node classes which reads better.

app/models/assignment.rb
jmileham: Ok, that ended up being tricky and adding a whole nother level of expressiveness to `AppRemoteKill.affecting`. Will do another review to outline that behavior.

app/models/assignment.rb
```diff
@@ -37,13 +38,19 @@ class Assignment < ActiveRecord::Base
...
+      Arel::Nodes::Grouping.new(
+        arel_table[:force].eq(true)
+        .or(Split.arel_excluding_incomplete_features_for(app_build))
       )
     )
   end
 
+  scope :excluding_remote_kills_for, ->(app_build) do
+    joins(:split).where(
+      Split.arel_excluding_remote_kills_for(app_build, override: arel_table[:force], overridden_at: arel_table[:updated_at])
```
jmileham: In order to make force assignments only win over remote kills when they're newer than remote kills, I ended up having to plumb some attributes through to the underlying scopes. In the process I learned some things that I now believe about the nuance of how force assignments should work with respect to feature completions and remote kills.

I now believe that a force assignment should always win over feature incompletion, even if a feature complete flag is set later. By definition you've already turned on the unfinished feature as an alpha tester or internal population. There's no reason that marking a flag feature complete should disable your access to the partially completed feature on your device. In practice, you're in a population that's likely to upgrade a lot and it may not matter much because you'll probably be on the feature complete version post-haste, but I'd argue that it'd be more surprising as an alpha tester or developer to see the feature shut itself down for you on your device as a result of marking a future version as feature complete.

Remote kills, on the other hand, are likely to be deployed in response to crash-bugs or similarly bad news incidents. If I'm a developer, I probably don't want my app to keep blowing up after a remote kill is put in place - I want my app to respect the remote kill immediately, but give me the option to toggle the feature back on if I want to see the faulty behavior for debugging purposes.

So this behavior offers us that. force assignments always win over feature incompleteness, but they only win over remote kills updated before our most recent assignment update.

app/models/app_remote_kill.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  scope :affecting, ->(app_build, override: false, overridden_at: nil) do
+    where(
+      arel_table[:app_id].eq(app_build.app_id)
+      .and(arel_table[:first_bad_version].lteq(app_build.version))
+      .and(arel_fixed_version_is_null_or_greater_than(app_build.version))
+      .and(
+        Arel::Nodes::Grouping.new(
+          arel_is_false(override)
+          .or(arel_table[:updated_at].gt(overridden_at))
```
jmileham: The idea of the `override` and `overridden_at` args to affecting is that we'll consider a given remote kill not to affect us if override evaluates to true and overridden_at is more recent than the remote kill's last modification.

But the cool thing is that they can take ruby value objects like `false`, `nil`, or a `TimeWithZone`, or they can take arel expressions, meaning we can have other scopes call us with join columns accessible via the outer context of a subselect. So that's what we do here. We can test it simply here, but leverage it for sweet multi-table where conditions in calling scopes.

app/models/app_remote_kill.rb
```diff
@@ -12,14 +12,28 @@ class AppRemoteKill < ActiveRecord::Base
...
+        )
+      )
     )
   end
 
+  class << self
+    private
+
+    def arel_is_false(value_or_arel)
+      Arel::Nodes::Grouping.new(Arel::Nodes::False.new).eq(value_or_arel)
```
jmileham: This turns out to be kind of annoying. Arel::Nodes::False doesn't mix in predications, and therefor can't be chained off of with `eq`, so I do this teensy hack and wrap it in a group node (parens) which does.

This way we get the automatic casting behavior of eq, which allows us to get a ruby value object or an arel node and work fine regardless.

jmileham:
@smudge << domain platform

app/models/app_remote_kill.rb
jmileham: @samandmoore the convo you had about remote kill being valuable for developer prerelease builds that want to make breaking API changes without revving the API version numbers is making me think that remote kills should not respect hard assignments at all regardless of recency. It would elevate remote kills to un-overrideable even by devs. Relative to the safety of not confusing an internal tester who might flip the flag on again and then see wildly broken app state, I’m now pretty bearish on allowing overrides to remote kills. If you want to see around a remote kill, you can cut yourself a local build with the flag hardcoded.

Aligns better with the seeming finality of a word like kill too (though kills can be edited/removed, they’d apply brutally when in place).

jmileham: Also the override behavior in local can be looser than on TT server once we have TT fake server to alleviate the potential pain of a dev trying to work on a bug fix for a feature that has an open-ended remote kill and potentially accidentally committing the gate for a not-yet-100%-weighted feature hard coded open.

jmileham: Given the way these PRs are stacked, I’m gonna suggest we review this code as it lies regardless and add the behavior change in a subsequent PR, and leave an architecture decision record about it (I introduced the repo’s first ADR in a subsequent PR too).

app/models/app_remote_kill.rb
```diff
@@ -0,0 +1,77 @@
+class AppRemoteKill < ActiveRecord::Base
+  belongs_to :app
+  belongs_to :split
+
+  attribute :first_bad_version, :app_version
+  attribute :fixed_version, :app_version
+
+  validates :app, :split, :reason, :override_to, :first_bad_version, presence: true
+  validates :reason, uniqueness: { scope: %i(app split) }
```
smudge: We can't ship and fix the same mistake twice?

app/models/app_remote_kill.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  def self.arel_fixed_version_is_null_or_greater_than(version)
+    Arel::Nodes::Grouping.new(
+      arel_table[:fixed_version].eq(nil)
+      .or(arel_table[:fixed_version].gt(version))
+    )
+  end
+
+  def override_to_must_exist
+    return unless split
```
smudge: Is this to prevent nil object errors during `valid?`? Any reason not to do `validate :override_to_must_exist, if: :split` at the top?

app/models/app_remote_kill.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    errors.add(:override_to, "must exist in split's current variants") unless split.has_variant?(override_to)
+  end
+
+  def fixed_version_must_be_greater_than_first_bad_version
+    return if fixed_version.nil?
```
smudge: same -- `validate :fixed_version_must_be_greater_than_first_bad_version, if: :fixed_version`?

app/models/app_remote_kill.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    errors.add(:fixed_version, "must be greater than first bad version") unless fixed_version > first_bad_version
+  end
+
+  def must_not_overlap_existing
+    return if first_bad_version.nil? || split.nil? || app.nil?
+
+    overlapping = self.class.overlapping(self)
+    return if overlapping.empty?
```
smudge: Hmm, so we're leaving "guard clause" territory... 🤔 Curious if you have strong opinions reasons to return early here.

app/models/app_remote_kill.rb
```diff
@@ -0,0 +1,77 @@
...
+  def fixed_version_must_be_greater_than_first_bad_version
+    return if fixed_version.nil?
+
+    errors.add(:fixed_version, "must be greater than first bad version") unless fixed_version > first_bad_version
+  end
+
+  def must_not_overlap_existing
+    return if first_bad_version.nil? || split.nil? || app.nil?
+
+    overlapping = self.class.overlapping(self)
```
smudge: Thoughts on pulling this out?

```ruby
def overlapping
  self.class.overlapping(self)
end
```

app/models/assignment.rb
smudge: Strongly agreeing with paragraph two. Paragraph 3 seems reasonable.

app/models/split.rb
```diff
@@ -48,12 +49,33 @@ class Split < ActiveRecord::Base
...
+        .exists
+      )
     )
   end
 
+  scope :with_remote_kill_knockouts_for, ->(app_build) do
+    previous_selects = all.arel.projections
+    except(:select)
+      .select(
+        previous_selects,
```
smudge: does `all.arel.projections` get mutated by the two lines above, or is the variable simplify for naming clarity?

db/migrate/20190409231451_create_app_remote_kills.rb
```diff
@@ -0,0 +1,15 @@
+class CreateAppRemoteKills < ActiveRecord::Migration[5.1]
+  def change
+    create_table :app_remote_kills do |t|
+      t.references :app, type: :uuid, null: false, foreign_key: true
+      t.references :split, type: :uuid, null: false, foreign_key: true
+      t.string "reason", null: false
+      t.string "override_to", null: false
+      t.integer "first_bad_version", array: true, null: false
+      t.integer "fixed_version", array: true
```
smudge: I'm assuming it'll be pretty trivial to go back in and set a fixed version once it is fixed.

app/models/app_remote_kill.rb
smudge: I guess it only ever applies in an unpersisted context.

app/models/app_remote_kill.rb
jmileham: the reason exists primarily as a human identifier for mutating the existing remote kill record.

app/models/app_remote_kill.rb
jmileham: this is basically like a validator that has an `allow_nil: true` or `allow_blank: true` variant, which is the flavor I usually prefer for all those validations because I use the presence validator on top.

app/models/app_remote_kill.rb
jmileham: As above, you could do it that way, but I see this as a complete unit - a validator that operates only on extant values.

app/models/app_remote_kill.rb
jmileham: I think it was threading the needle of line-too-long on an `unless overlapping.empty?` on the subsequent line. I don't love it

app/models/app_remote_kill.rb
jmileham: the validation does run on persisted stuff - there's a where clause to allow IDs to match. I don't feel strongly about it but it's not a mega useful method outside of this context. I'd want to cache it if we put it in an instance var cause i use it twice rapid fire, and I'm not sure about the cache invalidation story then.

app/models/split.rb
jmileham: it does get mutated, yep.

db/migrate/20190409231451_create_app_remote_kills.rb
jmileham: yeah, that's what the reason is for - you post up another migration for the same split app and reason and it sets fixed_version.

app/models/app_remote_kill.rb
```diff
@@ -0,0 +1,77 @@
...
+  attribute :fixed_version, :app_version
+
+  validates :app, :split, :reason, :override_to, :first_bad_version, presence: true
+  validates :reason, uniqueness: { scope: %i(app split) }
+
+  validate :override_to_must_exist
+  validate :fixed_version_must_be_greater_than_first_bad_version
+  validate :must_not_overlap_existing
+
+  scope :affecting, ->(app_build, override: false, overridden_at: nil) do
```
smudge: Taking a stab at what this basically translates to in SQL:

```
WHERE app_id = {{app_id}}
  AND first_bad_version <= {{version}}
  AND fixed_version IS NULL OR fixed_version > {{version}}
  AND {{"assignments.force" or true/false}} = false OR updated_at > {{"assignments.updated_at" or NULL}}
```

I'm wondering if you could save the query engine some work if you get non-Arel nodes for arguments. If `override` is true and `overridden_at` is `nil`, you can essentially change this whole thing to `WHERE false = true`, no? Maybe not worth it?

app/models/app_remote_kill.rb
jmileham: yeah, come to think about it, i wouldn't be comfortable memoizing  an `#overlapping` method, so I'd still be doing `local_overlapping = overlapping` or something weird here.

app/models/app_remote_kill.rb
jmileham: override = true and overriden_at is null is actually an illegal state, you either get both or neither. I needed them to be expressable as distinct arel nodes, though.

Also this is coming out in #112

jmileham: also the query engine will resolve literal-based logic to `WHERE false = true` before it does a single index scan, so i don't think we'd be interested in doing that logic in ruby instead of PG anyway

app/models/app_remote_kill.rb
smudge: Got it. So the structure is basically:

```ruby
def thing_must_behave
  return if something # nil-check

  errors.add(field, "must behave") if condition # actual validation condition
```

And the point is to keep the nil-check separate from the actual validation condition?

smudge: I'm not opposed to this as a choice, but I do wish the existing validations DSL / errors API made this a little easier to express without early `return`s everywhere.

app/models/app_remote_kill.rb
smudge: Right, so is it that we want to make sure that you're actually changing the reason each time (instead of writing "oops" or something every time) -- i.e. a tiny reminder that you're writing this text value for other humans? I'm wondering if we would also want to enforce a minimum length too.

jmileham: oops would probably work if you were breaking a different feature each time. Yeah, I dunno, I hope that with decent example text we'll get good results. Here's what I'm drafting in the CLI docs now:

`Reason should be a camel_case slug, e.g. "home_screen_crash_jan_2019".`

app/models/app_remote_kill.rb
jmileham: yeah, agreed.

app/models/split.rb
```diff
@@ -48,12 +49,33 @@ class Split < ActiveRecord::Base
...
+        AppRemoteKill.select(:override_to).affecting(app_build).arel
```
smudge: So the magic here is that you can rip out an `arel` node, adjust it a little, and then shove it straight into `.select()`? Fancy.

app/models/split.rb
```diff
@@ -112,7 +134,9 @@ def create_decision!(params = {})
   end
 
   def registry
-    if try(:feature_incomplete?) # This is a virtual attribute provided by the with_feature_incomplete_knockouts_for scope
+    if try(:remote_kill_override_to) # These are virtual attributes provided by the for_app_build scope
+      knock_out_weightings(super, to: remote_kill_override_to)
+    elsif try(:feature_incomplete?) # These are virtual attributes provided by the for_app_build scope
       knock_out_weightings(super)
```
smudge: I could see a very different world in which you somehow managed to do a `COALESCE(override_to, 'false')` somewhere in a very different set of queries. But in order to get to that place, you'd need to wrap some kind of polymorphic blanket around `FeatureCompletion`s and `RemoteKill`s. And while they are conceptually similar (both providing ranges over which assignments should/should not apply under certain conditions -- the distinction of future assignments overriding vs not overriding being one of the bigger differences) you'd have to solve the exact kind of overlapping prioritization problem that we are avoiding with `RemoteKill`s. So, not saying we should try to go there, but at the same time at the back of my brain there's a little engineer who wants a simpler / more elegant solution where perhaps there is nothing truly viable.

jmileham: sometimes you just wanna express a prioritized set of conditional actions. :) I'm saying here explicitly that remote kill override_to values will win over feature incomplete `false` if they are to different variants, which is actually a nice thing to see on the page, I think.

jmileham: (because both of these things can be true at once, so it should be clear to a reader that the remote_kill is going to take precedence)

app/models/app_remote_kill.rb
smudge: Ah, got it.

smudge:
<< domain && platform LGTM -- The functionality checks out, and I want to move onto the other PRs and maybe we can try to unify/wrap-up a few things at the end. But mainly -- and I'm sure I'm not the only one -- I think could benefit from some comments that provide example SQL that all of the Arel fragments might generate. (It's like trying to read a math paper with formulas fully written out in English. 😅)

smudge:
domain LGTM && platform LGTM?

jmileham:
yeah, sorry, the generated SQL here is nontrivial. :) Thanks!

PR #108, opened 2019-04-13
Feature completion migration endpoint

jmileham:
### Summary

This adds a new migration endpoint to create/update/destroy feature completions. I renamed a the AppFeatureCompletion.split_id column to feature_gate_id in the process and validated that it's the right kind of split now that I'm shoring this up for data integrity while receiving untrusted input.

I'll drop a comment with the version range that's real because this includes `remote_kill`'s commits right now.

### DIFF
--- a/app/models/app_feature_completion_migration.rb
+++ b/app/models/app_feature_completion_migration.rb
@@ -1,0 +1,57 @@
+class AppFeatureCompletionMigration
+  include ActiveModel::Validations
+
+  attr_reader :app, :feature_gate, :version_before_type_cast
+
+  validate :feature_gate_must_exist
+
+  delegate :version, to: :app_feature_completion
+
+  def initialize(params)
+    @app = params[:app] || raise("Must provide app")
+    @feature_gate = params[:feature_gate]
+    app_feature_completion.version = @version_before_type_cast = params[:version]
+  end
+
+  def valid?
+    if version_before_type_cast
+      app_feature_completion.valid?
+    else
+      super
+    end
+  end
+
+  def errors
+    if version_before_type_cast
+      app_feature_completion.errors
+    else
+      super
+    end
+  end
+
+  def save
+    if version_before_type_cast
+      app_feature_completion.save
+    else
+      if valid? && app_feature_completion.persisted?
+        app_feature_completion.destroy.destroyed?
+      else
+        false
+      end
+    end
+  end
+
+  def app_feature_completion
+    @app_feature_completion ||= app.feature_completions.find_or_initialize_by(feature_gate: feature_gate_model)
+  end
+
+  def feature_gate_model
+    Split.find_by(name: feature_gate)
+  end
+
+  private
+
+  def feature_gate_must_exist
+    errors.add(:feature_gate, "must exist") unless feature_gate_model
+  end
+end
--- a/app/models/app_feature_completion_migration.rb
+++ b/app/models/app_feature_completion_migration.rb
@@ -14,32 +14,34 @@ class AppFeatureCompletionMigration
   end
 
   def valid?
-    if version_before_type_cast
-      app_feature_completion.valid?
-    else
+    if destroy?
       super
+    else
+      app_feature_completion.valid?
     end
   end
 
   def errors
-    if version_before_type_cast
-      app_feature_completion.errors
-    else
+    if destroy?
       super
+    else
+      app_feature_completion.errors
     end
   end
 
   def save
-    if version_before_type_cast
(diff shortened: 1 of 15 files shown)


jmileham:
here are the real commits of this PR: https://github.com/Betterment/test_track/pull/108/files/6bf010c96827fefdacfcb30e7396659c312d2352..930bde2547451d1a76f7df1d447202b9004d2d9d

app/types/app_version_type.rb
```diff
@@ -6,7 +6,7 @@ def cast(value)
   end
 
   def serialize(value)
-    value.to_pg_array if value
+    AppVersion.new(value).to_pg_array if value
```
jmileham: Turns out that find_or_initialize_by doesn't flow through cast before hitting serialize. Luckily AppVersion initializes just fine from an AppVersion so no big deal.

jmileham:
@smudge << domain platform

Remember the diff that's actually distinct here is just: https://github.com/Betterment/test_track/pull/108/files/6bf010c96827fefdacfcb30e7396659c312d2352..fab1a02bc5d6ffcac5e60d5cf65b82a3c4af77e0

app/models/app_feature_completion_migration.rb
```diff
@@ -0,0 +1,57 @@
...
+  attr_reader :app, :feature_gate, :version_before_type_cast
+
+  validate :feature_gate_must_exist
+
+  delegate :version, to: :app_feature_completion
+
+  def initialize(params)
+    @app = params[:app] || raise("Must provide app")
+    @feature_gate = params[:feature_gate]
+    app_feature_completion.version = @version_before_type_cast = params[:version]
```
smudge: I'm having trouble figuring out why you can't `delegate :version_before_type_cast` too. Mind illuminating?

jmileham: i spose I could, I just would have to have confidence in my understanding of the nuances of the implementation of ARBTC, when I really don't need a nuanced implementation here - I was looking for a revealing way to express that I care about whether I was handed a blank one of these or not regardless of what happens in my downstream.

app/models/app_feature_completion_migration.rb
```diff
@@ -0,0 +1,57 @@
+class AppFeatureCompletionMigration
+  include ActiveModel::Validations
+
+  attr_reader :app, :feature_gate, :version_before_type_cast
+
+  validate :feature_gate_must_exist
+
+  delegate :version, to: :app_feature_completion
+
+  def initialize(params)
```
smudge: Any reason not to do kwargs here?

jmileham: yeah, they're hostile to StrongParameters, I believe? This thing is an immutable resource model effectively.

app/models/app_feature_completion_migration.rb
```diff
@@ -0,0 +1,57 @@
...
+    if destroy?
+      super
+    else
+      app_feature_completion.errors
+    end
+  end
+
+  def save
+    if destroy?
+      valid? && app_feature_completion.destroy.destroyed?
```
smudge: is `destroy` not sufficiently truthy/falsey?

app/models/app_feature_completion_migration.rb
smudge: Fair. I think I'd trust it to fit the needs of this class, especially if the tests accurately describe the behavior in question ("i pass an empty version string") without stubbing the BTC value or anything. But I buy that this puts it front and center for someone trying to figure out how the destroy case gets triggered. I might in that case rename it to `@destroy` and read directly from that within `destroy?` (or something to avoid name conflicts on attr names -- `destroy_record`?) to make that a much tighter loop.

app/models/app_feature_completion_migration.rb
jmileham: This might be a mental tic, but destroy returns the model (which is why `destroyed?` is a thing). I'm not certain it would ever not return true from `destroyed?` but it seems like it'd be interesting if it didn't, and we should investigate, so I make it a 422.

app/models/app_feature_completion_migration.rb
jmileham: I'm trying to reference some common language about what's going on by naming it the same thing that AR would name it. I can't just set `@destroy` to the BTC value, I'd need to call .present? on the BTC value to make it behave correctly, and right now that resides in the destroy method, making it more than just a delegation.

smudge:
domain LGTM && platform LGTM 👍

PR #109, opened 2019-04-13
Remote kill migration endpoint

jmileham:
### Summary

just like #108 only for remote_kills.

Will call out relevant diff below.

### DIFF
--- a/app/models/app_remote_kill.rb
+++ b/app/models/app_remote_kill.rb
@@ -10,7 +10,6 @@ class AppRemoteKill < ActiveRecord::Base
     uniqueness: { scope: %i(app split) },
     format: { with: /\A[a-z\d_]*\z/, message: "must be alphanumeric snake_case" }
 
-  validate :override_to_must_exist
   validate :fixed_version_must_be_greater_than_first_bad_version
   validate :must_not_overlap_existing
 
@@ -54,12 +53,6 @@ class AppRemoteKill < ActiveRecord::Base
       arel_table[:fixed_version].eq(nil)
       .or(arel_table[:fixed_version].gt(version))
     )
-  end
-
-  def override_to_must_exist
-    return unless split
-
-    errors.add(:override_to, "must exist in split's current variants") unless split.has_variant?(override_to)
   end
 
   def fixed_version_must_be_greater_than_first_bad_version
--- a/app/models/app_remote_kill.rb
+++ b/app/models/app_remote_kill.rb
@@ -6,7 +6,9 @@ class AppRemoteKill < ActiveRecord::Base
   attribute :fixed_version, :app_version
 
   validates :app, :split, :reason, :override_to, :first_bad_version, presence: true
-  validates :reason, uniqueness: { scope: %i(app split) }
+  validates :reason,
+    uniqueness: { scope: %i(app split) },
+    format: { with: /\A[a-z\d_]*\z/, message: "must be alphanumeric snake_case" }
 
   validate :override_to_must_exist
   validate :fixed_version_must_be_greater_than_first_bad_version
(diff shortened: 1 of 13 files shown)


jmileham:
Relevant commits: https://github.com/Betterment/test_track/pull/109/files/fab1a02bc5d6ffcac5e60d5cf65b82a3c4af77e0..8b418fde3693175f0db6351c4d59c218100dd42e

spec/models/split_spec.rb
```diff
@@ -443,7 +443,7 @@
       end
 
       it "overrides to 100% remote_kill_override_to if remote killed" do
-        fc = FactoryBot.create(:app_remote_kill, split: subject, first_bad_version: "0.9", fixed_version: "1.1")
+        fc = FactoryBot.create(:app_remote_kill, split: subject, override_to: "touch_this", first_bad_version: "0.9", fixed_version: "1.1")
```
jmileham: I apparently had some mystery guest testing going on where I was asserting on the factory default. woops. :)

jmileham:
@smudge << domain platform

Relevant commits again: https://github.com/Betterment/test_track/pull/109/files/fab1a02bc5d6ffcac5e60d5cf65b82a3c4af77e0..074e56494272deb2f218c9ca5423eed39f8324d8

app/models/app_remote_kill.rb
```diff
@@ -6,9 +6,10 @@ class AppRemoteKill < ActiveRecord::Base
   attribute :fixed_version, :app_version
 
   validates :app, :split, :reason, :override_to, :first_bad_version, presence: true
-  validates :reason, uniqueness: { scope: %i(app split) }
+  validates :reason,
+    uniqueness: { scope: %i(app split) },
+    format: { with: /\A[a-z\d_]*\z/, message: "must be alphanumeric snake_case" }
```
smudge: 👍

app/models/app_remote_kill_migration.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def save
+    if destroy?
+      valid? && app_remote_kill.destroy.destroyed?
+    else
+      app_remote_kill.save
+    end
+  end
+
+  def destroy?
```
smudge: Does this need to be public? (Same for feature completions)

doc/arch/adr-001.md
```diff
@@ -0,0 +1,44 @@
...
+Accepted
+
+## Consequences
+
+The only remaining way to ship an AppRemoteKill that doesn't apply
+correctly in production would be to misspell the split name itself.
+
+We'll see whether that is acceptable or whether we should walk further
+down the decoupling path and make AppRemoteKills and
+AppFeatureCompletions entirely soft-coupled to splits by name.
```
smudge: In theory, this would allow us to preemptively kill a split before it even exists, right?

spec/models/app_remote_kill_migration_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+      app: app,
+      split: feature_gate.name,
+      reason: "",
+      override_to: "false",
+      first_bad_version: nil,
+      fixed_version: nil
+    )
+
+    expect(subject).to have(1).error_on(:reason)
+    expect(subject.save).to eq false
```
smudge: Worth checking that this doesn't change the row count?

smudge:
domain LGTM && platform LGTM

app/models/app_remote_kill_migration.rb
jmileham: it doesn't need to be, but my general take is that if I'm not embarrassed by it being depended on by a collaborator or test, I will tend to go public. Turns out I didn't test it directly, but I wouldn't have felt bad doing that.

doc/arch/adr-001.md
jmileham: yeah, totally, which might actually be good depending on release cadence.

spec/models/app_remote_kill_migration_spec.rb
jmileham: will do!

PR #110, opened 2019-04-13
App migrations

jmileham:
### Summary

This provides persistence/retrieval for migration versions so clients can know which migrations they still need to run, even when they merge out of order.

Will follow with relevant commits.

### DIFF
--- a/db/migrate/20190413205758_create_app_migrations.rb
+++ b/db/migrate/20190413205758_create_app_migrations.rb
@@ -5,6 +5,6 @@ class CreateAppMigrations < ActiveRecord::Migration[5.1]
       t.string :version, null: false, unique: true
     end
 
-    add_index :app_migrations, :version, unique: true
+    add_index :app_migrations, [:app_id, :version], unique: true
   end
 end
--- a/db/migrate/20190413205758_create_app_migrations.rb
+++ b/db/migrate/20190413205758_create_app_migrations.rb
@@ -1,0 +1,10 @@
+class CreateAppMigrations < ActiveRecord::Migration[5.1]
+  def change
+    create_table :app_migrations, id: :uuid, default: -> { "uuid_generate_v4()" } do |t|
+      t.references :app, type: :uuid, null: :false, foreign_key: true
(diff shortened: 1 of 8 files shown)


jmileham:
Here's the relevant diff: https://github.com/Betterment/test_track/pull/110/files/8fb319521483dd2bddeac93d5f534540b8ea333f..fbff778493683c24593e7c3720176f9946cd7f61

jmileham:
@samandmoore << domain platform

db/migrate/20190413205758_create_app_migrations.rb
```diff
@@ -0,0 +1,10 @@
+class CreateAppMigrations < ActiveRecord::Migration[5.1]
+  def change
+    create_table :app_migrations, id: :uuid, default: -> { "uuid_generate_v4()" } do |t|
+      t.references :app, type: :uuid, null: :false, foreign_key: true
+      t.string :version, null: false, unique: true
```
samandmoore: `unique: true` doesn't appear to be doing anything. is that okay for you?

samandmoore: seems like you don't need it here since you have the specific index below.

samandmoore:
<<domain platform

samandmoore:
domainlgtm platformlgtm!

db/migrate/20190413205758_create_app_migrations.rb
samandmoore: @jmileham might just wanna remove this to avoid confusion

smudge: If you wanted it to work, would it be `index: { unique: true }`?

jmileham: yeah, i will remove - it is too tight, wouldn't allow multiple apps to have migrations with the same timestamp, which is unnecessary.

jmileham:
```suggestion
      t.string :version, null: false
```

PR #111, opened 2019-04-15
Mobile client endpoints

jmileham:
### Summary

Provides endpoints for clients in the field to filter assignments and do contextual variant knockouts.

I believe this will be the last TT PR before MVP of mobile extensions is done server side.

### Other Information

Per usual there's a bunch of commits in here that are not relevant due to stacked PRs. I'll call out the real diff below.

/domain @JonLz for mobile integration needs - this would be what you'd be implementing to client side.

### DIFF
--- a/app/controllers/api/v1/app_identifier_visitor_configs_controller.rb
+++ b/app/controllers/api/v1/app_identifier_visitor_configs_controller.rb
@@ -1,0 +1,21 @@
+class Api::V1::AppIdentifierVisitorConfigsController < UnauthenticatedApiController
+  include CorsSupport
+
+  def show
+    app_build = AppVersionBuildPath.new(build_params).app_build
+    @active_splits = Split.for_presentation(app_build: app_build)
+    visitor = VisitorLookup.new(identifier_params).visitor
+    @visitor = visitor.id
+    @assignments = visitor.assignments_for(app_build).includes(:split).order(:updated_at)
+  end
+
+  private
+
+  def build_params
+    params.permit(:app_name, :version_number, :build_timestamp)
+  end
+
+  def identifier_params
+    params.permit(:identifier_type_name, :identifier_value)
+  end
+end
--- a/app/controllers/api/v1/app_identifier_visitor_configs_controller.rb
+++ b/app/controllers/api/v1/app_identifier_visitor_configs_controller.rb
@@ -5,7 +5,7 @@ class Api::V1::AppIdentifierVisitorConfigsController < UnauthenticatedApiControl
     app_build = AppVersionBuildPath.new(build_params).app_build
     @active_splits = Split.for_presentation(app_build: app_build)
     visitor = VisitorLookup.new(identifier_params).visitor
-    @visitor = visitor.id
+    @visitor_id = visitor.id
     @assignments = visitor.assignments_for(app_build).includes(:split).order(:updated_at)
   end
 
--- a/app/controllers/api/v1/app_identifier_visitor_configs_controller.rb
+++ b/app/controllers/api/v1/app_identifier_visitor_configs_controller.rb
@@ -1,12 +1,17 @@
 class Api::V1::AppIdentifierVisitorConfigsController < UnauthenticatedApiController
   include CorsSupport
 
-  def show
-    app_build = AppVersionBuildPath.new(build_params).app_build
-    @active_splits = Split.for_presentation(app_build: app_build)
-    visitor = VisitorLookup.new(identifier_params).visitor
-    @visitor_id = visitor.id
-    @assignments = visitor.assignments_for(app_build).includes(:split).order(:updated_at)
+  def show # rubocop:disable Metrics/AbcSize
+    build_path = AppVersionBuildPath.new(build_params)
+    if build_path.valid?
+      app_build = AppVersionBuildPath.new(build_params).app_build
+      @active_splits = Split.for_presentation(app_build: app_build)
+      visitor = VisitorLookup.new(identifier_params).visitor
+      @visitor_id = visitor.id
+      @assignments = visitor.assignments_for(app_build).includes(:split).order(:updated_at)
+    else
+      render_errors build_path
+    end
   end
 
   private
--- a/app/models/app.rb
+++ b/app/models/app.rb
@@ -28,7 +28,7 @@ class App < ActiveRecord::Base
     def initialize(app_id:, version:, built_at:)
       @app_id = app_id
       @version = AppVersion.new(version)
-      @built_at = built_at
+      @built_at = built_at.is_a?(String) ? Time.zone.parse(built_at) : built_at
     end
   end
 end
--- a/app/models/app_version_build_path.rb
+++ b/app/models/app_version_build_path.rb
@@ -1,19 +1,46 @@
 class AppVersionBuildPath
-  attr_reader :app_name, :version, :built_at
+  include ActiveModel::Validations
+
+  attr_reader :app_name, :version_number, :build_timestamp
+
+  validates :app_name, :version_number, :build_timestamp, presence: true
+  # We want to make sure the client is shipping us high-precision ISO
+  # timestamps so we choose to allow timestamps at either millisecond or second
+  # precision based on the W3C interpretation of iso8601 from this SO answer:
+  #
+  # https://stackoverflow.com/a/3143231
+  validates :build_timestamp, format: {
+    with: /\A\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d(.\d+)?([+-][0-2]\d:[0-5]\d|Z)\z/, allow_blank: true
+  }
+  validate :version_number_must_be_parseable
 
   def initialize(params)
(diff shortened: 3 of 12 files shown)


jmileham:
Here's the relevant diff: https://github.com/Betterment/test_track/pull/111/commits/73f198813fd3a01b227992e673aa065277578dd2

app/models/app.rb
```diff
@@ -28,7 +28,7 @@ class Build
     def initialize(app_id:, version:, built_at:)
       @app_id = app_id
       @version = AppVersion.new(version)
-      @built_at = built_at
+      @built_at = built_at.is_a?(String) ? Time.zone.parse(built_at) : built_at
```
jmileham: Note that we'll parse anything that's parseable, but clients should definitely choose to represent timestamp in something unambiguous with a time zone.

config/routes.rb
```diff
@@ -24,6 +24,21 @@
         end
       end
 
+      resources :apps, only: [], param: :name do
+        resources :versions, only: [], param: :number, constraints: { number: /[\d\.]+/ } do
+          resources :builds, only: [], param: :timestamp do
+            resources :visitors, only: [] do
+              resource :config, only: :show, controller: 'app_visitor_configs'
```
jmileham: Here's what this looks like as a path (extracted from rake routes):

`/api/v1/apps/:app_name/versions/:version_number/builds/:build_timestamp/visitors/:visitor_id/config`

config/routes.rb
```diff
@@ -24,6 +24,21 @@
...
+            end
+            resources :identifier_types, only: [], param: :name do
+              resources :identifiers, only: [], param: :value do
+                resource :visitor_config, only: :show, controller: 'app_identifier_visitor_configs'
```
jmileham: and this one looks like:

`/api/v1/apps/:app_name/versions/:version_number/builds/:build_timestamp/identifier_types/:identifier_type_name/identifiers/:identifier_value/visitor_config`

Notably, even though we've established a different format for v2 splits and assignments endpoints, to make the lift smaller for mobile, I've built these as v1-style controllers.

app/views/api/v1/app_visitor_configs/_show.json.jbuilder
```diff
@@ -0,0 +1,10 @@
+json.splits do
+  active_splits.each do |split|
+    json.set! split.name, split.registry
+  end
+end
+
+json.visitor do
+  json.id visitor_id
+  json.assignments assignments, :split_name, :variant, :context, :unsynced
+end
```
jmileham: These look like this on the wire:

```json
{
  "splits": {
    "bla_enabled": { "false": 100, "true": 0 },
    "other_enabled": { "false": 100, "true": 0 },
    "an_experiment": { "control": 50, "treatment": 50 }
  },
  "visitor": {
    "id": "some_uuid",
    "assignments": [
      {
        "split_name": "bla_enabled",
        "variant": "true",
        "context": "wherever",
        "unsynced": false
      },
      {
        "split_name": "an_experiment",
        "variant": "treatment",
        "context": "somewhereelse",
        "unsynced": false
      }
    ]
  }
}
```

config/routes.rb
jmileham: You'll definitely want to make sure your URL-interpolated values are URLencoded properly - especially the build timestamp.

jmileham:
@smudget << domain platform

Relevant commits: [there is only one and it messes it up if i show it](https://github.com/Betterment/test_track/pull/111/commits/73f198813fd3a01b227992e673aa065277578dd2)

jmileham:
nanda

config/routes.rb
samandmoore: by `v1-style` do you mean, within the `v1` namespace? if so, okay!

samandmoore:
@smudge >> domain platform
<<domain platform

config/routes.rb
jmileham: it's in the v1 namespace, but it also conforms to the way the v1 visitors controller renders, not the way the v2 visitors controller renders, which nobody uses yet.

spec/controllers/api/v1/app_visitor_configs_controller_spec.rb
```diff
@@ -0,0 +1,62 @@
+require 'rails_helper'
+
+RSpec.describe Api::V1::AppVisitorConfigsController do
+  describe "#show" do
+    let(:app) { FactoryBot.create(:app) }
+    let(:identifier_type) { FactoryBot.create(:identifier_type, name: "clown_id") }
```
samandmoore: you can delete this line here, nit

app/models/app.rb
samandmoore: we don't wanna be more prescriptive here and say it's gotta be an iso timestamp?

app/controllers/api/v1/app_identifier_visitor_configs_controller.rb
```diff
@@ -0,0 +1,21 @@
+class Api::V1::AppIdentifierVisitorConfigsController < UnauthenticatedApiController
+  include CorsSupport
+
+  def show
+    app_build = AppVersionBuildPath.new(build_params).app_build
+    @active_splits = Split.for_presentation(app_build: app_build)
+    visitor = VisitorLookup.new(identifier_params).visitor
+    @visitor = visitor.id
```
samandmoore: is this supposed to be `@visitor_id` like in the other controller?

samandmoore: seems like the tests should be failing

samandmoore:
domaintafn platformlgtm -- but i think you've got one typo in there and i can't understand why the test isn't failing...

app/views/api/v1/app_identifier_visitor_configs/show.json.jbuilder
```diff
@@ -0,0 +1,4 @@
+json.partial! 'api/v1/app_visitor_configs/show',
+  active_splits: @active_splits,
+  visitor_id: @visitor_id,
```
samandmoore: seems like this line should be blowing up because we don't assign @visitor_id

jmileham:
ivars fre

app/views/api/v1/app_identifier_visitor_configs/show.json.jbuilder
jmileham: yeah, that's just classic "unset ivars return nil" behavior. womp. I fixed and added a test granular enough to catch it.

samandmoore: yea, sorry, i just meant that i was surprised it didn't get caught in a test. coolcoolcool

samandmoore:
domainlgtm

app/models/app.rb
jmileham: Yeah I dunno, I could, but i don’t have a great precedent for sending 422s back in response to a get. I guess I can still do it and it’ll get the point across to client implementors and fail safer.

jmileham: Note that iso dates don’t have to include minutes and seconds, so maybe I need to do a regex for a canonical form with the precision I want to see.

app/views/api/v1/app_identifier_visitor_configs/show.json.jbuilder
jmileham: I don’t assert on the whole response because there’s a fair bit of factory defaults in there. It’s unusual to ship an ID solo like this, but I wanted to keep the assignment fetch explicit and in the controller (so nobody accidentally grabs visitor.assignments in the view layer and hurts themselves). That and the identifier controller was a copy paste of the visitor controller which was structurally immune to this issue since the @visitor_id was literally used to fetch assignments. so I didn’t intuitively test it. :(

JonLz:
<< domain lgtm

jmileham:
@samandmoore - I went to town on validating the URL params. You now must bring an ISO date with second or millisecond precision or we'll return you a 422 with an error bag, which should be helpful to client library devs. Notably an app that can't be found by name will 404, not 422.

Here's the new commit: https://github.com/Betterment/test_track/pull/111/commits/c47dd9d17706c8094d3728eda54a03d5d2983af2

Re-LG please? It's a substantial diff.

app/models/app_version_build_path.rb
```diff
@@ -1,19 +1,46 @@
...
     App.find_by!(name: app_name)
   end
+
+  def version_number_must_be_parseable
+    return if version_number.blank?
+
+    AppVersion.new(version_number)
+  rescue StandardError
+    errors.add(:version_number, <<~ERROR)
+      must be a valid in alignment with https://github.com/Betterment/test_track/blob/master/app/models/app_version.rb
```
samandmoore: must be a valid ___ ? or just `must be in alignment with`?

jmileham:
```suggestion
      must be valid according to https://github.com/Betterment/test_track/blob/master/app/models/app_version.rb
```

samandmoore:
small comment, otherwise domainlgtm

PR #113, opened 2019-04-23
Add V2 splits controller expecting app name prefix

jmileham:
### Summary

Add a split endpoint for the new migration runner to hook up to. This now requires app prefixes. See ADR-003 for details


jmileham:
Oh yeah, I need to not have a destroy endpoint because retirement is going to be a thing. Hold please.

doc/arch/adr-003.md
```diff
@@ -0,0 +1,51 @@
...
+not have the latest version of each app at all times and a
+fully-upgraded local testtrack server to detect conflicts that might
+arise in production.
+
+We need to make sure that migrations don't get jammed up in production
+even if a developer only has their own app downloaded locally.
+
+## Decision
+
+New migration runners will be expected to prefix their split names with
```
smudge: Maybe we're too far down one path for this to make sense, but I'd be interested in your thoughts on allowing apps to have a local scope on split names, where they don't need to set their own prefix (or even see it on lookups). They'd only see prefixes for splits owned by _other_ apps.

jmileham:
Yup! That is exactly where we’re headed! The wire protocol will be fully
qualified, but the migrations and client variant/AB calls will default to
local scope.

doc/arch/adr-003.md
smudge: > Yup! That is exactly where we’re headed! The wire protocol will be fully qualified, but the migrations and client variant/AB calls will default to local scope.

Oh, nice. Then carry on 😉

app/models/split.rb
```diff
@@ -209,6 +228,17 @@ def name_ends_with_test?
...
 
+  def name_prefixed_other_than_app_name?
+    parts = name.split(".")
+    if parts.length == 1 && !require_app_name_prefix?
+      return false
+    elsif parts.length == 2
+      return parts.first != owner_app.name
+    else
+      return true
+    end
```
samandmoore: forget you were writing ruby here? i think you can remove all the `return`s

jmileham:
bumpity bump

app/controllers/api/v1/identifier_types_controller.rb
```diff
@@ -1,6 +1,6 @@
 class Api::V1::IdentifierTypesController < AuthenticatedApiController
   def create
-    identifier_type = IdentifierType.create_with(owner_app: current_app).find_or_initialize_by(create_params)
+    identifier_type = IdentifierType.find_or_initialize_by(create_params.merge(owner_app: current_app))
```
smudge: Only thing here is that it's gonna start subtly acting wonky if someone adds additional, non-uniquely-identifying params to `create_params`. Maybe this can be expressed like so?:

```ruby
identifier_type = IdentifierType.find_or_initialize_by(owner_app: current_app, name: create_params[:name])
```

smudge: But I guess it's not really different from the way it was expressed before, so take it or leave it. 🤷‍♂️

app/controllers/api/v2/migrations/splits_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+    else
+      render_errors split_upsert
+    end
+  end
+
+  private
+
+  def create_params
+    # rails-sponsored workaround https://github.com/rails/rails/pull/12609
+    params.permit(:name, weighting_registry: params[:weighting_registry].try(:keys))
```
smudge: Oh interesting. This doesn't work?:

```ruby
params.permit(:name, weighting_registry: {})
```

smudge:
```ruby
pry(main)> ActionController::Parameters.new(test: { a: 2, c: { banana: 'foo' } }, fish: 1).permit(test: {})
=> <ActionController::Parameters {"test"=><ActionController::Parameters {"a"=>2, "c"=><ActionController::Parameters {"banana"=>"foo"} permitted: true>} permitted: true>} permitted: true>

pry(main)> _.to_h
=> {"test"=>{"a"=>2, "c"=>{"banana"=>"foo"}}}

Comment (jmileham) [reply app/controllers/api/v2/migrations/splits_controller.rb]: I dunno, I copied this from the previous working controller and would be hesitant to change it unless you've got confidence it won't break anything. :) It may have been to work around a limitation of rails that no longer exists?

File: app/controllers/api/v1/identifier_types_controller.rb (hunk 3 above)
Comment (jmileham) [reply app/controllers/api/v1/identifier_types_controller.rb]: Yeah, in this case, I made it slightly tighter than the previous iteration in that it won't tell you it's OK if you try to touch another app's identifier_type. This definitely isn't as resilient to future change of this controller as it could be (if I wanted that I'd probably build a migration model for it), but it's also the least used feature of testtrack. Important, but mostly only used when you first set it up for an app ecosystem.

File: app/models/split.rb [hunk 5]
```diff
@@ -143,6 +144,18 @@ def registry
     end
   end
 
+  def require_app_name_prefix=(value)
+    if instance_variable_defined?(:@require_app_name_prefix)
+      @require_app_name_prefix
+    else
+      @require_app_name_prefix = ActiveRecord::Type::Boolean.new.cast(value)
+    end
+  end
```
smudge: I think you can be fancy with the new APIs here:

```ruby
attribute : require_app_name_prefix, :boolean
```

Should give you the predicate method for free too.

jmileham: There are several instances of doing it this way in the code already, so I'd probably want to change them all, which I sort of didn't want to uncork :)

app/models/split.rb
```diff
@@ -143,6 +144,18 @@ def registry
     end
   end
 
+  def require_app_name_prefix=(value)
+    if instance_variable_defined?(:@require_app_name_prefix)
+      @require_app_name_prefix
```
smudge: In case my other suggestion doesn't work, the setter doesn't need to return anything here - it will already return the passed value by default.

jmileham: oh right

app/controllers/api/v2/migrations/splits_controller.rb
smudge: As long as the `weighting_registry` is constructed of permitted scalars (i.e. string, numbers, dates, etc) it should be fine. It can be deeply nested, contain lists, etc -- the `{}` argument is meant as an escape hatch for exactly this kind of use case. Maybe it historically wasn't so permissive? I'd be curious to see what the test cases say.

smudge: Oh I see, maybe in this case it explicitly _shouldn't_ contain any nesting, despite how the keys are arbitrary? That makes more sense. Maybe the comment could mention that.

smudge: Doesn't quite feel like strong params is the best way to guard for that -- once you're out of "known params" territory into the land of arbitrary/dynamic params, something at the model layer could validate that the values are all numbers, etc.

app/models/split.rb
smudge: As long as it's in your mental backlog 🤷‍♂️

app/controllers/api/v2/migrations/splits_controller.rb
jmileham: Yeah, jogging my memory, I think it’s that, but I’m general the strong params model’s job is to ensure you don’t get funny shaped data, not that the scalars you get in the correct spots are invalid. Then it’s a validation and serialization concern. So yeah, I could comment it, but I’m not convinced strongparams isn’t doing exactly what I want it to here.

app/models/split.rb
smudge: I _will_ call out one of my favorite takeaways from last year's RailsConf:
https://twitter.com/sarahmei/status/1084980571019796481

But then I'll move onto the rest of the PR 😉

jmileham: I think I’d likely only change if they deprecated this API :) my brain is not large enough to fit a refactoring backlog in it. :)

app/models/split.rb
```diff
@@ -198,7 +217,7 @@ def registry_weights_must_be_integers
   end
 
   def name_not_underscored?
-    name && !underscored?(name)
+    name && !underscored?(name.split('.').last)
```
smudge: Do we validate that the app name is snake cased too?

smudge: (not seeing it on `app.rb`)

app/models/split_retirement_migration.rb
```diff
@@ -0,0 +1,40 @@
...
+  def split_model
+    @split_model ||= Split.find_by(owner_app: app, name: split)
+  end
+
+  def split_must_exist
+    errors.add(:split, "must exist and belong to app") unless split_model
+  end
+
+  def decision_must_exist
+    errors.add(:decision, "must exist in split") unless split_model&.has_variant?(decision)
```
smudge: Minor, but this validation will add an error if `split_model` is nil when it probably doesn't need to. (i mean, maybe? depends on how you look at it.)

app/models/split.rb
jmileham: Actually I wanted that more flexible so app name can match any repo name which often has dashes, depending on your platform persuasion. That relaxation is actually why I made this change.

spec/models/split_retirement_migration_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  it "is invalid with a missing split" do
+    app = FactoryBot.create(:app)
+    expect(described_class.new(app: app, split: "my_split", decision: "b")).to have(1).error_on(:split)
+  end
+
+  it "is invalid with a split from the wrong app" do
+    FactoryBot.create(:split, name: "my_split", registry: { a: 100, b: 0 })
+    other_app = FactoryBot.create(:app)
+
+    expect(described_class.new(app: other_app, split: "my_split", decision: "b")).to have(1).error_on(:split)
```
smudge: (this is where you could theoretically check for 0 errors on `:decision`)

smudge: (and above)

app/models/split_retirement_migration.rb
jmileham: Yeah, seems like I should tighten because it could be a red herring with a typo in the split name but a totally legit variant

spec/models/split_spec.rb
```diff
@@ -47,6 +49,43 @@
...
+      expect(subject.errors[:name].first).to include("can only be prefixed with")
+    end
+
+    it "is fine with new splits when require_app_name_prefix is set to something false-ish" do
+      subject.name = 'baz'
+      subject.require_app_name_prefix = "f"
+      expect(subject).to be_valid
+    end
+
+    it "is fine with dashes in app names" do
```
smudge: got it 👍

smudge:
cool,

<< domain LGTM && platform LGTM

just the 1 validation adjustment and the return value of the setter, but those are minor and don't really impact what's happening overall here.

PR #114, opened 2019-04-25
Add split decisions controller

jmileham:
### Summary

Adds a new migration endpoint for deciding splits from the new migration CLI. It's very similar to retirement except it actually unretires a split that has been retired. This expressive capability to be able to unretire a split back to a decided state is why it makes sense to even make this a migration, otherwise, it would be impossible to revert a retirement of a decided split without temporarily pivoting through active and clicking a button in the admin console.

### DIFF
--- a/app/models/split_decision_migration.rb
+++ b/app/models/split_decision_migration.rb
@@ -1,0 +1,39 @@
+class SplitDecisionMigration
+  include ActiveModel::Validations
+
+  attr_reader :app,
+    :split,
+    :variant
+
+  validate :split_must_exist
+  validate :variant_must_exist
+
+  def initialize(params)
+    @app = params[:app] || raise("Must provide app")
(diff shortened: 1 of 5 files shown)


app/models/split_decision_migration.rb
```diff
@@ -0,0 +1,39 @@
...
+  def split_model
+    @split_model ||= Split.find_by(owner_app: app, name: split)
+  end
+
+  def split_must_exist
+    errors.add(:split, "must exist and belong to app") unless split_model
+  end
+
+  def variant_must_exist
+    errors.add(:variant, "must exist in split") unless split_model&.has_variant?(variant)
```
smudge: Similar here if `split_model` is nil.

smudge:
<< domain LGTM && platform LGTM

PR #115, opened 2019-05-05
Don't merge feature gate assignments when superseding visitors

jmileham:
### Summary

Don't copy feature_gate assignments on visitor supersession to make it impossible to piggyback on privileged identities to get into closed features.

### Other Information

Also added a test to codify the behavior that assignments to decided splits should either not be copied or their updated_at be backdated so as to hide the assignment. The current implementation doesn't copy them, and that is fine. There's a story for copying them and backdating updated_at, but I don't think it's that important. It would only be interesting if a split were undecided and we wanted the assignment to come back to life for the superseding identity, but that feels unnecessary. Either way the current test will work if that behavior is changed.

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -51,6 +51,8 @@ class Assignment < ActiveRecord::Base
     )
   end
 
+  scope :except_feature_gates, -> { joins(:split).merge(Split.except_feature_gates) }
+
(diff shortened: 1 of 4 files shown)


samandmoore:
Works for me. 

<<domainlgtm platformlgtm

PR #116, opened 2019-05-05
Jmileham/relax naming validations

jmileham:
### Summary

now that testtrack-cli runs headless, we really can't have server side validations that invalidate otherwise good splits. The test suffix is no longer possible, and the outlaw of the word `new` was kind of an opinion that really doesn't hold up that well for a company based in `new_york`. We'll have to find other ways to encourage good split naming.


samandmoore:
<<domainlgtm platformlgtm

PR #117, opened 2019-06-19
Add sentry support, make airbrake optional

jmileham:
### Summary

Adds sentry support and makes Airbrake support optional (not requiring the gem if not needed).

/platform @samandmoore @aburgel

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,12 @@
 language: ruby
 rvm:
-- 2.4.2
-- 2.6.2
+  - 2.4.2
+  - 2.6.2
 addons:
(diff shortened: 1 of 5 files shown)


aburgel:
<< domain platform

the switch to bundler 2 is causing the ruby 2.3.1 build to fail. i think its time to put ruby 2.3 to bed.

jmileham:
Actually 2.3 would've been fine with bundler 2.0.1 if I'd done the travis config change I just did (I think) but I'm happy having newer ruby instead anyway.

aburgel:
domainlgtm platformlgtm

PR #119, opened 2019-07-08
Touch assignments to ensure chrome extension sees changes

aburgel:
/no-platform

This addresses a UX bug in the chrome extension.

For a decided, but still active, split where a previous assignment existed, using the chrome extension to assign the same variant effectively no-ops, which looks like nothing happens within the chrome extension. This makes it appear as if test track is broken.

How we get here is a little convoluted, but here are the important points:

1. Assigning to the same variant results in no change in the assignments table, including no change to the `updated_at` attribute.

2. For a decided split, assignments that were updated before the split's `decided_at` will not appear in the assignment registry.

3. Within the chrome extension, a split appears unassigned if there is no record in the assignment registry.

Using the chrome extension, if the user attempts to create an assignment when the database has an existing assignment for that variant, the updated_at attribute of the assignment will remain the unchanged, and the assignment will continue to appear "unassigned".

To fix, I'm always updating the `updated_at` regardless of whether the assignment actually changed. This should more closely match the user's expectation that using the chrome extension should "force" assignment changes.

### DIFF
--- a/app/models/arbitrary_assignment_creation.rb
+++ b/app/models/arbitrary_assignment_creation.rb
@@ -1,8 +1,7 @@
 class ArbitraryAssignmentCreation
   attr_reader :visitor_id, :split_name, :variant, :bulk_assignment_id, :context, :force
 
-  # rubocop:disable Metrics/ParameterLists
-  def initialize(
+  def initialize( # rubocop:disable Metrics/ParameterLists
     visitor_id: nil,
     split_name: nil,
     variant: nil,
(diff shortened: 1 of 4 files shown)


app/models/arbitrary_assignment_creation.rb
```diff
@@ -43,10 +41,10 @@ def save_new_assignment!
     save!
   end
 
-  def supersede!(now = Time.zone.now)
```
jmileham: was this optional arg just never used?

aburgel: yeah, i couldn't find any use. i'm guessing it was meant for documentation purposes.

jmileham:
<< domain lgtm!

PR #120, opened 2019-07-09
upgrade omniauth-saml

jmileham:
### Summary

Security patch noise reduction for omniauth-saml to 1.10.1 which now requires ruby-saml >= 1.7 which fixed CVE-2017-11430, even though we already had ruby-saml 1.9 in our Gemfile.lock, so we were already patched up.

/no-platform


jmileham:
Also hit CVE-2015-9284. This is a pretty tricky one because omniauth gem decided not to accept a fix (optionally) relying on rails CSRF protection in order to mitigate. They suggested creating an omniauth-rails gem which never got released. Subsequently Cookpad released a different gem that patches things up similar to the original omniauth proposal.

So now we're good if we added multi-SAML-provider support. I'm not sure how to get github alerts to be quiet up about our vulnerable version of omniauth now that we've mitigated. But lets get this in first.

smudge:
<< domain LGTM

PR #123, opened 2019-07-16
Bugfix: Noop assignment_events for existing assignments

jmileham:
### Summary

#119 introduced a change that unintentionally caused the `assignment_event` endpoint to touch updated_at on existing assignments for decided splits when clients, seeing no assignment (due to assignment masking in the relatively-new decision logic) attempted to create fresh assignments. As a result, the old assignment would win over the decision, effectively negating the decision.

This only impacts experiments, as feature gates are already immune to being explicitly assigned via the `assignment_event` endpoint

### Other Information

I got to this implementation by testing from the outside in and redefining a test whose title matched what I wanted to be more strict. Ultimately it turns out that the `DeterministicAssignmentCreation` model was way fancier than it needed to be, unless I'm missing something.

DeterministicAssignmentCreation is only used by the AssignmentEvents controller, so the blast radius of this change, even if it is buggy, is contained. There's already a relatively nasty bug in play here, so my gut is to move aggressively here.

/platform @coreyja @samandmoore

### DIFF
--- a/app/models/deterministic_assignment_creation.rb
+++ b/app/models/deterministic_assignment_creation.rb
@@ -34,7 +34,7 @@ class DeterministicAssignmentCreation
   end
 
   def existing_assignment
-    @existing_assignment ||= Assignment.find_by visitor: visitor, split: split
+    Assignment.find_by visitor: visitor, split: split
   end
 
   def visitor
--- a/app/models/deterministic_assignment_creation.rb
+++ b/app/models/deterministic_assignment_creation.rb
@@ -14,19 +14,15 @@ class DeterministicAssignmentCreation
   end
 
   def save!
-    unless split.feature_gate?
+    if !split.feature_gate? && !existing_assignment
       ArbitraryAssignmentCreation.create!(
         visitor_id: visitor_id,
         split_name: split_name,
-        variant: variant,
+        variant: variant_calculator.variant,
         mixpanel_result: mixpanel_result,
         context: context
       )
     end
-  end
-
-  def variant
-    assignment.variant || variant_calculator.variant
   end
 
   def variant_calculator
@@ -37,8 +33,8 @@ class DeterministicAssignmentCreation
     @split ||= Split.find_by!(name: split_name)
   end
 
-  def assignment
-    @assignment ||= Assignment.find_or_initialize_by visitor: visitor, split: split
+  def existing_assignment
+    @existing_assignment ||= Assignment.find_by visitor: visitor, split: split
   end
 
   def visitor
--- a/spec/controllers/api/v1/assignment_events_controller_spec.rb
+++ b/spec/controllers/api/v1/assignment_events_controller_spec.rb
@@ -30,8 +30,12 @@ RSpec.describe Api::V1::AssignmentEventsController, type: :controller do
     end
 
     it "noops if a conflicting assignment already exists" do
-      assignment = FactoryBot.create(:assignment, visitor: visitor, split: split, variant: "control")
+      assignment = nil
+      Timecop.freeze(1.year.ago) do
+        assignment = FactoryBot.create(:assignment, visitor: visitor, split: split, variant: "control")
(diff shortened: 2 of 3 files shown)


spec/models/deterministic_assignment_creation_spec.rb
```diff
@@ -55,13 +55,12 @@
...
 
-    it "recreates with the previously assigned variant if already assigned" do
+    it "noops if already assigned" do
       FactoryBot.create(:assignment, split: split, visitor: Visitor.from_id("bc8833fd-1bdc-4751-a13c-8aba0ef95a3b"), variant: "variant3")
 
       subject.save!
 
-      expect(ArbitraryAssignmentCreation).to have_received(:create!)
-        .with(hash_including(variant: "variant3"))
+      expect(ArbitraryAssignmentCreation).not_to have_received(:create!)
```
jmileham: I'm very much changing tested behavior here, but the only exercise codepath is thru assignment_events, which I can't see how this behavior would provide value to.

jmileham:
Another option here would've been to,  additionally or instead, specifically noop assignment_events post decision, but that doesn't feel quite right. While it's true that the end-user-visible bug manifests only for decided splits with existing assignments to the non-decision variant, it actually feels OK for a first-time assignment to a decided split to be recorded. This would lead to UX continuity for a visitor who first experienced the split post-decision, should the decision be unmade (e.g. because it was called early or by accident).

At the same time, it doesn't feel appropriate to ever touch updated_at when receiving a duplicate assignment_event for a visitor, so that's what this PR proposes.

spec/controllers/api/v1/assignment_events_controller_spec.rb
```diff
@@ -30,13 +30,20 @@
...
+      Timecop.freeze(1.year.ago) do
+        assignment = FactoryBot.create(:assignment, visitor: visitor, split: split, variant: "control")
+      end
+
+      assignment.reload
+      original_attributes = assignment.attributes
 
       expect {
         post :create, params: create_params
       }.not_to change { PreviousAssignment.count }
```
coreyja: We might be able to extend this to also test that we don't change the assignment attributes. Something like:
```
expect {
  post :create, params: create_params
}.not_to change { PreviousAssignment.count }
.and not_change { assignment.reload.attributes }
```

Though that might also mean you would need to create the negated `not_change` matcher, assuming the TT code base doesn't already have it.

So take it or leave it! Obviously not a big deal! On my first read I just missed why we were saving the `original_attributes`

jmileham: seems a little fancy. I'm OK with not understanding why you need original_attributes until they're checked.

app/models/deterministic_assignment_creation.rb
```diff
@@ -37,8 +33,8 @@ def split
     @split ||= Split.find_by!(name: split_name)
   end
 
-  def assignment
-    @assignment ||= Assignment.find_or_initialize_by visitor: visitor, split: split
+  def existing_assignment
+    @existing_assignment ||= Assignment.find_by visitor: visitor, split: split
```
coreyja: We might want to drop the memoization from here. I _think_ we only call `existing_assignment` once now, and since we can't actually memoize the `nil` I wonder if this is just going to cause slight dev confusion going forward

jmileham: sure
```suggestion
    Assignment.find_by visitor: visitor, split: split
```

coreyja:
<< platform LGTM

Also happy to grab domain if you want, but leaving it off since I am slightly less familiar with the TT modeling and stuff. BUT I am also relatively familiar with this bug now, so happy to grab domain too if you want!

coreyja:
<< domain LGTM

I just got indoctured as a new member of `test_track_core`! So seems like I might as well use my new powers to give this an LG

PR #124, opened 2019-07-17
Show decided state in admin

### LINKED ISSUE #122: Admin does not display decided state of splits
Decision is a part of a splits state now, and it's not reflected in the admin panel at all.

/cc @maidoesthings @coreyja

jmileham:
### Summary

fixes #122

### Other Information

There are no tests on this field, and I didn't add any, but the Active codepath actually covers the finished and decided conditionals via fallthru, so this would be hard to break unless the definitions of finished or decided changed.

### DIFF
--- a/app/models/split.rb
+++ b/app/models/split.rb
@@ -157,10 +157,7 @@ class Split < ActiveRecord::Base
   end
 
   def decided_variant
-    return nil unless decided?
-    registry.each_with_object([]) { |(k, v), found|
-      found << k if v == 100
-    }.first
+    registry.key(100) if decided?
   end
 
   private
--- a/app/models/split.rb
+++ b/app/models/split.rb
@@ -101,6 +101,10 @@ class Split < ActiveRecord::Base
     finished_at.present?
   end
 
+  def decided?
+    decided_at.present?
+  end
+
   def reassign_weight(weighting_registry)
     now = Time.zone.now
     previous_split_registries.build(registry: registry, created_at: updated_at, updated_at: now, superseded_at: now)
--- a/app/models/split.rb
+++ b/app/models/split.rb
@@ -156,6 +156,13 @@ class Split < ActiveRecord::Base
     @require_app_name_prefix
   end
 
+  def decided_variant
+    return nil unless decided?
+    registry.each_with_object([]) { |(k, v), found|
+      found << k if v == 100
+    }.first
+  end
+
   private
 
   def knock_out_weightings(registry_hash, to: "false")
--- a/app/views/admin/splits/_details.html.erb
+++ b/app/views/admin/splits/_details.html.erb
@@ -19,7 +19,13 @@
       </tr>
       <tr>
         <td>Status</td>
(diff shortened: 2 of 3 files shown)


app/views/admin/splits/_details.html.erb
```diff
@@ -19,7 +19,13 @@
       </tr>
       <tr>
         <td>Status</td>
-        <td class="u-status--<%= @split.finished? ? 'finished' : 'active' %>"><%= @split.finished? ? 'Finished' : 'Active' %><span class="sc-p--s">🏁</span></td>
+        <% if @split.finished? %>
+               <td class="u-status--finished">Finished (retired)<span class="sc-p--s">🏁</span></td>
+        <% elsif @split.decided? %>
+               <td class="u-status--decided">Decided</td>
```
coreyja: Would it be beneficial/easy to add what variant it was decided to here?

coreyja:
<< platform LGTM

<< domain TAFN

Code looks great! Left one small idea about adding the variant we decided too

app/views/admin/splits/_details.html.erb
```diff
@@ -19,7 +19,13 @@
       </tr>
       <tr>
         <td>Status</td>
-        <td class="u-status--<%= @split.finished? ? 'finished' : 'active' %>"><%= @split.finished? ? 'Finished' : 'Active' %><span class="sc-p--s">🏁</span></td>
+        <% if @split.finished? %>
+               <td class="u-status--finished">Finished (retired)<span class="sc-p--s">🏁</span></td>
```
smudge: I'll be the one to nitpick on using two-space tabs in ERB templates.

jmileham: fine :)

jmileham:
Good call. Screenies below

Decided:

Screen Shot 2019-07-17 at 12 41 03 PM

Finished:

Screen Shot 2019-07-17 at 12 42 29 PM

jmileham:
bump

app/models/split.rb
```diff
@@ -152,6 +156,13 @@ def require_app_name_prefix?
     @require_app_name_prefix
   end
 
+  def decided_variant
+    return nil unless decided?
+    registry.each_with_object([]) { |(k, v), found|
+      found << k if v == 100
+    }.first
```
smudge: Could this just become `registry.key(100)`?

jmileham: yes, didn't know or forgot that was a thing!

jmileham:
bumpalump

spec/models/split_spec.rb
```diff
@@ -452,6 +452,25 @@
...
+
+      it "returns nil if decided and not 100% weighted" do
+        subject = FactoryBot.build_stubbed(:split, registry: { false: 50, true: 50 }, decided_at: Time.zone.now)
+
+        expect(subject.decided_variant).to be_nil
+      end
+
+      it "returns 100% weighted variant if decided" do
+        subject = FactoryBot.build_stubbed(:split, registry: { false: 0, true: 100 }, decided_at: Time.zone.now)
+        expect(subject.decided_variant).to eq("true")
```
jmileham:
```suggestion

        expect(subject.decided_variant).to eq("true")
```

coreyja:
<< domain LGTM

PR #125, opened 2019-08-08
Update assignment when mixpanel result has changed

joejansen:
### Summary

https://github.com/Betterment/test_track/pull/123 introduced a change that results in the `mixpanel_result` never being updated when the client sends a request to `DeterministicAssignmentCreation` for an existing assignment. I believe the sequence of events would look something like this:

1. The client encounters a split and sends the first request to `Api::V1::AssignmentEventsController` with params that have a `nil` `mixpanel_result`.
2. The client's request to the analytics library succeeds and it sends another request to `Api::V1::AssignmentEventsController` with params that have a `"success"` `mixpanel_result`.
3. The `DeterministicAssignmentCreation` doesn't call `ArbitraryAssignmentCreation` because it finds an `existing_assignment`.
4. On a subsequent page load the client retrieves the split registry and sees a split with `"unsynced": true`.
5. The client sends another request to `Api::V1::AssignmentEventsController` with params that have a `"success"` `mixpanel_result`.
6. The `DeterministicAssignmentCreation` doesn't call `ArbitraryAssignmentCreation` because it finds an `existing_assignment`.

4 to 6 keep repeating for each page load.

This changes proposes a `should_create_or_update_assignment?` expression that incorporates a comparison of whether the existing and new `mixpanel_result`s are equal.

It also adds an optional `created_at` attr to `ArbitraryAssignmentCreation` so we can choose to preserve the existing assignment timestamp if we don't want the default `now`.

/platform @jmileham @aburgel

### DIFF
--- a/app/models/deterministic_assignment_creation.rb
+++ b/app/models/deterministic_assignment_creation.rb
@@ -14,13 +14,14 @@ class DeterministicAssignmentCreation
   end
 
   def save!
-    if !split.feature_gate? && !existing_assignment
+    if should_create_or_update_assignment?
       ArbitraryAssignmentCreation.create!(
         visitor_id: visitor_id,
         split_name: split_name,
         variant: variant_calculator.variant,
         mixpanel_result: mixpanel_result,
-        context: context
+        context: context,
+        updated_at: updated_at
       )
     end
   end
@@ -39,5 +40,15 @@ class DeterministicAssignmentCreation
 
   def visitor
     @visitor ||= Visitor.from_id(visitor_id)
+  end
+
+  private
+
+  def should_create_or_update_assignment?
+    !split.feature_gate? && (!existing_assignment || existing_assignment.mixpanel_result != @mixpanel_result)
+  end
+
+  def updated_at
+    existing_assignment && existing_assignment.updated_at
   end
 end
--- a/spec/controllers/api/v1/assignment_events_controller_spec.rb
+++ b/spec/controllers/api/v1/assignment_events_controller_spec.rb
@@ -29,10 +29,10 @@ RSpec.describe Api::V1::AssignmentEventsController, type: :controller do
       expect(assignment.context).to eq "context"
     end
 
-    it "noops if a conflicting assignment already exists" do
+    it "noops if a conflicting assignment with a mixpanel_result already exists" do
       assignment = nil
       Timecop.freeze(1.year.ago) do
-        assignment = FactoryBot.create(:assignment, visitor: visitor, split: split, variant: "control")
+        assignment = FactoryBot.create(:assignment, visitor: visitor, split: split, variant: "control", mixpanel_result: "success")
       end
 
       assignment.reload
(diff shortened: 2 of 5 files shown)


aburgel:
<< domain

app/models/deterministic_assignment_creation.rb
```diff
@@ -14,13 +14,14 @@ def self.create!(params)
...
-    if !split.feature_gate? && !existing_assignment
+    if should_create_or_update_assignment?
       ArbitraryAssignmentCreation.create!(
         visitor_id: visitor_id,
         split_name: split_name,
         variant: variant_calculator.variant,
         mixpanel_result: mixpanel_result,
-        context: context
+        context: context,
+        updated_at: updated_at
```
aburgel: Some of this coupling between `DeterministicAssignmentCreation` and `ArbitraryAssignmentCreation` makes this code hard to follow.

Since the behavior around `mixpanel_result` is specific to existing assignments, I think we can increase the clarity by isolating the changes to this class.

What we want to express is that an assignment event that occurs when there already is an assignment means that we are only updating the `mixpanel_result`.

Something like:

```ruby
if existing_assignment
  existing_assignment.assign_attributes(mixpanel_result: mixpanel_result)
  existing_assignment.save!(touch: false)
else
  ArbitraryAssignmentCreation...
end
```

I realize we paired on this, but seeing it in PR form makes me think we're introducing too much complexity.

spec/controllers/api/v1/assignment_events_controller_spec.rb
```diff
@@ -46,6 +46,23 @@
...
+      assignment = nil
+      Timecop.freeze(1.year.ago) do
+        assignment = FactoryBot.create(:assignment, visitor: visitor, split: split, variant: "control", mixpanel_result: nil)
+      end
+
+      assignment.reload
+
+      expect {
+        post :create, params: create_params
+      }.to change { PreviousAssignment.count }.by(1)
```
aburgel: It's surprising that `PreviousAssignment` would change. That means we're switching variants, which means that in the case of a changed `mixpanel_result`, we could move someone into a new variant. That shouldn't happen if they already have an assignment.

In the normal case, the window would be pretty small since the second assignment event call with the updated mixpanel result should occur very quickly. But that's not guaranteed.

I think this points to us not reusing `ArbitraryAssignmentCreation` just to update the `mixpanel_result`.

spec/models/deterministic_assignment_creation_spec.rb
```diff
@@ -55,12 +55,34 @@
...
 
-      subject.save!
+          subject.save!
+
+          expect(ArbitraryAssignmentCreation).not_to have_received(:create!)
+        end
+      end
+
+      context 'when the mixpanel result is different' do
+        let(:date) { Date.parse('2016-08-07') }
```
aburgel: Let's use `Time.zone.parse` since `updated_at` is a timestamp.

jmileham:
<< platform - Burgel's gonna take this over while @joejansen's out

jmileham:
@aburgel >> domain
<< domain

cuz why not

jmileham:
tafn

jmileham:
nanda

aburgel:
Reopened with some modification in #126 since this PR off a fork.

PR #126, opened 2019-08-19
Update assignment when mixpanel result has changed

aburgel:
NOTE: This is built upon #125. I had to open a new PR since that PR was off of a fork. I made the following modification: Consolidated the logic in `DeterministicAssignmentCreation` to reduce coupling to `ArbitraryAssignmentCreation` and to close the window in which a new variant could  be unintentially assigned.

### Summary

https://github.com/Betterment/test_track/pull/123 introduced a change that results in the `mixpanel_result` never being updated when the client sends a request to `DeterministicAssignmentCreation` for an existing assignment. I believe the sequence of events would look something like this:

1. The client encounters a split and sends the first request to `Api::V1::AssignmentEventsController` with params that have a `nil` `mixpanel_result`.
2. The client's request to the analytics library succeeds and it sends another request to `Api::V1::AssignmentEventsController` with params that have a `"success"` `mixpanel_result`.
3. The `DeterministicAssignmentCreation` doesn't call `ArbitraryAssignmentCreation` because it finds an `existing_assignment`.
4. On a subsequent page load the client retrieves the split registry and sees a split with `"unsynced": true`.
5. The client sends another request to `Api::V1::AssignmentEventsController` with params that have a `"success"` `mixpanel_result`.
6. The `DeterministicAssignmentCreation` doesn't call `ArbitraryAssignmentCreation` because it finds an `existing_assignment`.

4 to 6 keep repeating for each page load.

~~This changes proposes a `should_create_or_update_assignment?` expression that incorporates a comparison of whether the existing and new `mixpanel_result`s are equal.~~

~~It also adds an optional `created_at` attr to `ArbitraryAssignmentCreation` so we can choose to preserve the existing assignment timestamp if we don't want the default `now`.~~

/domain @jmileham 
/platform @jmileham 
/cc @joejansen

### DIFF
--- a/app/models/arbitrary_assignment_creation.rb
+++ b/app/models/arbitrary_assignment_creation.rb
@@ -1,5 +1,5 @@
 class ArbitraryAssignmentCreation
-  attr_reader :visitor_id, :split_name, :variant, :bulk_assignment_id, :context, :force
+  attr_reader :visitor_id, :split_name, :variant, :bulk_assignment_id, :context, :force, :updated_at
 
   def initialize( # rubocop:disable Metrics/ParameterLists
     visitor_id: nil,
@@ -8,7 +8,8 @@ class ArbitraryAssignmentCreation
     mixpanel_result: nil,
     bulk_assignment_id: nil,
     context: nil,
-    force: false
+    force: false,
+    updated_at: nil
   )
     @visitor_id = visitor_id
     @split_name = split_name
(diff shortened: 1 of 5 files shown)


jmileham:
<< domain platform

jmileham:
should we add a ratchet so you can't change the mixpanel result after it's already in a terminal state? That's an optional feature add so domain lgtm platform lgtm!

aburgel:
> should we add a ratchet so you can't change the mixpanel result after it's already in a terminal state

i like this idea. i'm now only allowing `mixpanel_result` to change if its not 'success'

PR #136, opened 2019-12-19
Run tests against postgres 11

alederman:
Update CI to run against postgresql 11

### Summary
CI tests are run currently run against PG 9.6. This update will ensure TestTrack supports the latest major postgres version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,12 +14,16 @@ addons:
       - postgresql-11
       - postgresql-client-11
 
+env:
+  global:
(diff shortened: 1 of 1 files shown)


smudge:
@alederman as a temporary travis-specific workaround, could you add `gem uninstall bundler` a step before `gem update --system`? (gem update now ships a version of bundler that happens to conflict with the one pre-installed on travis, and it opens a (y/n) prompt that causes travis to hang).

PR #145, opened 2020-11-12
Add support for batch assignment overrides via web extension

samandmoore:
### Summary

It's common for users of the web extension to edit multiple split assignments at once. Up until now, this requires a separate API call for each assignment. This is inefficient, so we're introducing a new endpoint that can process a batch of assignment overrides all at once.

### Other Information

To take advantage of this, the web extension will need to be updated, and probably the JS client as well.

In the meantime, the old endpoint will keep on working 👍🏻 

/no-platform

~NOTE: i'm having issues installing libv8 + therubyracer locally. it might be caused by xcode 12, but i'm not sure. so there may be failing tests right now. i wrote them up but wasn't able to run them locally 😬~
I've got a workaround, use `mini_racer` locally 😄

### DIFF
--- a/app/controllers/api/v1/batch_assignment_overrides_controller.rb
+++ b/app/controllers/api/v1/batch_assignment_overrides_controller.rb
@@ -1,0 +1,14 @@
+class Api::V1::BatchAssignmentOverridesController < SharedSecretAuthenticatedApiController
+  include CorsSupport
+
+  def create
+    BatchArbitraryAssignmentCreation.create! create_params.merge(force: true)
+    head :no_content
+  end
+
+  private
+
+  def create_params
+    params.permit(:visitor_id, assignments: [:split_name, :variant, :mixpanel_result, :context])
+  end
+end
--- a/app/controllers/api/v1/batch_assignment_overrides_controller.rb
+++ b/app/controllers/api/v1/batch_assignment_overrides_controller.rb
@@ -9,6 +9,6 @@ class Api::V1::BatchAssignmentOverridesController < SharedSecretAuthenticatedApi
   private
 
   def create_params
-    params.permit(:visitor_id, assignments: [:split_name, :variant, :mixpanel_result, :context])
+    params.permit(:visitor_id, assignments: %i(split_name variant context))
   end
 end
--- a/app/controllers/api/v2/assignment_overrides_controller.rb
+++ b/app/controllers/api/v2/assignment_overrides_controller.rb
@@ -1,4 +1,4 @@
-class Api::V1::BatchAssignmentOverridesController < SharedSecretAuthenticatedApiController
+class Api::V2::AssignmentOverridesController < SharedSecretAuthenticatedApiController
   include CorsSupport
 
   def create
--- a/app/models/batch_arbitrary_assignment_creation.rb
+++ b/app/models/batch_arbitrary_assignment_creation.rb
@@ -1,0 +1,25 @@
+class BatchArbitraryAssignmentCreation
+  attr_reader :visitor_id, :assignments, :force
+
+  def initialize( # rubocop:disable Metrics/ParameterLists
+    visitor_id: nil,
+    assignments: nil,
+    force: false
+  )
+    @visitor_id = visitor_id
+    @assignments = assignments
+    @force = force
+  end
+
+  def save!
+    ActiveRecord::Base.transaction do
+      assignments.each do |assignment|
+        ArbitraryAssignmentCreation.create! assignment.merge(visitor_id: visitor_id, force: force)
+      end
+    end
+  end
+
+  def self.create!(params)
+    new(params.to_h.symbolize_keys).tap(&:save!)
+  end
+end
--- a/app/models/batch_arbitrary_assignment_creation.rb
+++ b/app/models/batch_arbitrary_assignment_creation.rb
@@ -2,8 +2,8 @@ class BatchArbitraryAssignmentCreation
   attr_reader :visitor_id, :assignments, :force
 
   def initialize(
-    visitor_id: nil,
-    assignments: nil,
+    visitor_id:,
+    assignments:,
     force: false
   )
     @visitor_id = visitor_id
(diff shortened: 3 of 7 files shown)


spec/models/batch_arbitrary_assignment_creation_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      expect(ArbitraryAssignmentCreation).to have_received(:create!).with(
+        visitor_id: visitor_id,
+        split_name: "split2",
+        variant: "variant2",
+        mixpanel_result: "success",
+        context: "the_context",
+        force: true
+      )
+    end
+  end
```
samandmoore, line 57: i decided that we really only need to test that it hands off to the other class because there's a robust suite of tests on that other class.

spec/controllers/api/v1/batch_assignment_overrides_controller_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+            assignment = Assignment.second
+            expect(assignment.variant).to eq "treatment"
+            expect(assignment.visitor).to eq visitor
+            expect(assignment.split).to eq split2
+            expect(assignment.mixpanel_result).to eq "success"
+            expect(assignment.context).to eq "context"
+            expect(assignment.force).to eq true
+          end
+        end
+      end
```
samandmoore, line 138: all the tests in this file _except for this one_ are duplicated from the controller tests for the singular assignment override endpoint. 

this one shows that the behavior is correct when there are multiple inputs

config/routes.rb
```diff
@@ -43,6 +43,10 @@
       match 'assignment_override', to: '/api/v1/cors#allow', via: :options
       resource :assignment_override, only: :create
 
+      # Shared secret-based batch assignment override for chrome extension
+      match 'batch_assignment_override', to: '/api/v1/cors#allow', via: :options
+      resource :batch_assignment_override, only: :create
```
samandmoore, line 48: i'm open to suggestions, but i opted to drop this right into the /v1 namespace alongside the existing endpoint because it's purely additive.

app/models/batch_arbitrary_assignment_creation.rb
```diff
@@ -0,0 +1,25 @@
...
+    @assignments = assignments
+    @force = force
+  end
+
+  def save!
+    ActiveRecord::Base.transaction do
+      assignments.each do |assignment|
+        ArbitraryAssignmentCreation.create! assignment.merge(visitor_id: visitor_id, force: force)
+      end
+    end
```
samandmoore, line 19: i _think_ that this new behavior is as simple as this.

app/controllers/api/v1/batch_assignment_overrides_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def create
+    BatchArbitraryAssignmentCreation.create! create_params.merge(force: true)
+    head :no_content
+  end
+
+  private
+
+  def create_params
+    params.permit(:visitor_id, assignments: [:split_name, :variant, :mixpanel_result, :context])
```
samandmoore, line 12: i left the `context` at the per-assignment-override level in the event that someone wants to provide that value per split.

if you think it should be top-level like the `visitor_id` i'm fine with that too.

also, i noticed that the original assignment endpoint has the `visitor_id` in the POST body and not in the url, so i kept that the same here. `POST /v1/batch_assignment_override`.

i'm open to naming suggestions. we already have a concept of a `BulkAssignmentCreation` which is focused on assigning a bunch of visitors to the same variant of a single split. whereas this code is assigning a single visitor a specific variant (per split) for a bunch of splits.

aburgel: deprecate the old endpoint and call this `V2:AssignmentOverridesController`? I don't see a need for the old one since this one can easily be swapped in.

i don't recall why we put visitor_id in the body. i do prefer the other way though, so if you take the suggestion of moving to v2, then i'd advocate for making that change too.

app/models/batch_arbitrary_assignment_creation.rb
```diff
@@ -0,0 +1,25 @@
+class BatchArbitraryAssignmentCreation
+  attr_reader :visitor_id, :assignments, :force
+
+  def initialize(
+    visitor_id: nil,
+    assignments: nil,
```
aburgel, line 6:
```suggestion
    visitor_id:,
    assignments:,
```

Any reason to leave these as optional?

app/controllers/api/v1/batch_assignment_overrides_controller.rb
aburgel, line 12: separately, we [always pass in 'success' as mixpanel_result](https://github.com/Betterment/test_track_js_client/blob/master/src/assignmentOverride.ts#L45).

we have a longer term goal of removing mixpanel-specific things from the backend, so i'd like to remove it from here as well. we can set the value in `BatchArbitraryAssignmentCreation` since I don't think we're fully ready to rip it out.

aburgel:
<< domain tafn!

app/models/batch_arbitrary_assignment_creation.rb
samandmoore: nope, i copied the `ArbitraryAssignmentCreation`, but it doesn't make sense to me that they're optional!

app/controllers/api/v1/batch_assignment_overrides_controller.rb
samandmoore: perfecto! will change.

app/controllers/api/v2/assignment_overrides_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def create
+    BatchArbitraryAssignmentCreation.create! create_params.merge(force: true)
+    head :no_content
+  end
+
+  private
+
+  def create_params
+    params.permit(:visitor_id, assignments: %i(split_name variant context))
```
samandmoore, line 12: `visitor_id` comes from the path, but this works all the same way

right now this is expecting the payload will be:

```
{
  assignments: [...]
}
```

sounds good?

samandmoore:
bump!

aburgel:
domainlgtm!

PR #146, opened 2020-11-30
update ddtrace to 0.43

kelvin-acosta:
### Summary

This PR updates the ddtrace gem by running `bundle update ddtrace --conservative`. This gets `test_track` up to date with the latest ddtrace release. It was previously using `0.17.2` and since then there have been many releases. 

I've also updated the datadog configuration to use environment variables for configuring the datadog agent host and the datadog trace agent port.


aburgel:
<< domainlgtm platformlgtm

config/initializers/datadog.rb
```diff
@@ -5,10 +5,25 @@
...
+                   analytics_enabled: true
+
       c.use :active_record, orm_service_name: "#{service_name}-active_record"
+
       c.use :delayed_job, service_name: "#{service_name}-delayed_job"
-      c.use :http, service_name: "#{service_name}-http", distributed_tracing: true
+
+      c.use :http, service_name: "#{service_name}-http",
+                   distributed_tracing: true,
+                   analytics_enabled: true
```
RowanMcDonald, line 26: Do we also want to add Faraday? 

``` ruby
    c.use :faraday, service_name: "#{Runtime.app_name}-faraday",
                    distributed_tracing: true
```

Got this from: https://github.com/Betterment/runtime/blob/main/config/initializers/datadog.rb#L39

kelvin-acosta: Yeah, I'll add that faraday config too

kelvin-acosta: Oh yeah we can, I'll add that too.

PR #147, opened 2021-01-08
Add Feature Completions and Remote Kills to Split details screen

samandmoore:
### Summary

Since we added "remote kills" and "feature completions" to the TestTrack feature set, we have not updated the Admin UI to display them. This PR adds both of these features as tables on the split details screen.

**feature gate with feature completion and remote kill**
image

**feature gate with no completions**
image

**experiments don't use feature completions**
image

**experiment with no completions and no remote kills**
image

### Other Information

I also upgraded to mini_racer because i couldn't run the app locally otherwise.

/no-platform


app/controllers/api/v2/migrations/split_retirements_controller.rb
```diff
@@ -4,6 +4,7 @@ def create
     if split_retirement.save
       head :no_content
     else
+      puts split_retirement.errors.full_messages
```
samandmoore, line 7:
```suggestion
```

app/controllers/api/v2/migrations/splits_controller.rb
```diff
@@ -4,6 +4,7 @@ def create
     if split_upsert.save
       head :no_content
     else
+      puts split_upsert.errors.full_messages
```
samandmoore, line 7:
```suggestion
```

app/controllers/api/v2/migrations/app_feature_completions_controller.rb
```diff
@@ -4,6 +4,7 @@ def create
     if feature_completion_migration.save
       head :no_content
     else
+      puts feature_completion_migration.errors.full_messages
```
samandmoore, line 7:
```suggestion
```

app/models/split.rb
```diff
@@ -6,6 +6,13 @@ class Split < ActiveRecord::Base
   has_many :bulk_assignments, dependent: :nullify
   has_many :variant_details, dependent: :nullify
 
+  has_many :remote_kills, ->(split) { where(split: split) }, through: :owner_app, class_name: 'AppRemoteKill', inverse_of: :split
+  has_many :feature_completions,
+    ->(split) { where(feature_gate: split) },
+    through: :owner_app,
+    class_name: 'AppFeatureCompletion',
+    inverse_of: :feature_gate
```
samandmoore, line 14: had to format this all weird so that it didn't trigger the line length cop

app/models/split.rb
```diff
@@ -6,6 +6,13 @@ class Split < ActiveRecord::Base
   has_many :bulk_assignments, dependent: :nullify
   has_many :variant_details, dependent: :nullify
 
+  has_many :remote_kills, ->(split) { where(split: split) }, through: :owner_app, class_name: 'AppRemoteKill', inverse_of: :split
```
aburgel, line 9: i don't follow this relation. why use `through: :owner_app`? remote_kill has a foreign key to split

app/models/split.rb
```diff
@@ -6,6 +6,13 @@ class Split < ActiveRecord::Base
...
+  has_many :feature_completions,
+    ->(split) { where(feature_gate: split) },
```
aburgel, line 11: same here, but in this case the foreign key has a non-conventional name: `feature_gate_id`

app/views/admin/splits/_feature_completions.html.erb
```diff
@@ -0,0 +1,40 @@
...
+              </tr>
+            <% end %>
+          </tbody>
+        </table>
+      <% else %>
+        <div class="EmptyTable EmptyTable--centered sc-p--l">
+          <p>
+            <strong>This feature has not been marked as complete for any app versions yet.</strong>
+            <br>
+            As a result, this feature is not visible to any clients that rely on feature completions (barring alpha tester overrides).
```
aburgel, line 28: i don't know the full rules of feature completions... is there something specific to `alphas` or non-final releases or do you just mean overrides in general?

app/views/admin/splits/_remote_kills.html.erb
```diff
@@ -0,0 +1,34 @@
...
+          <th>Reason</th>
+          <th>From version</th>
+          <th>To version</th>
+        </thead>
+        <tbody>
+          <% split.remote_kills.each do |remote_kill| %>
+            <tr>
+              <td><%= remote_kill.app.name %></td>
+              <td><%= remote_kill.reason %></td>
+              <td><%= remote_kill.first_bad_version %></td>
```
aburgel, line 20: `first_bad_version` is marked as [an arrary in the schema](https://github.com/Betterment/test_track/blob/master/db/schema.rb#L63:L63)... tbh, i'm not sure what that means, but do we need to handle it in some special way?

aburgel (commented):
<< domain tafn!

app/models/split.rb
samandmoore: Feature completions are per app, I believe remote kills are as well but I have to check on that. 

So in practice the feature could be complete for one app and not another.

app/views/admin/splits/_feature_completions.html.erb
samandmoore: Specifically overrides that are marked with the alpha checkbox in the admin UI

samandmoore: Although I think that web extension overrides also set that checkbox

app/views/admin/splits/_remote_kills.html.erb
samandmoore: Oh I think it's an array because it's a Semver value

samandmoore: ...but I don't think we have to handle it special because it's deserialized into a type called AppVersion

app/models/split.rb
aburgel: yeah, that makes sense, but i think that means we shouldn't use a `through`. both split and feature completions `belong_to` an `app`, if in practice, those are always the same app, then using a through shouldn't make a difference, it'll just be an extra unnecessary join. if they can be different, then the `through` will exclude some feature completions from different apps.

i guess it hinges on that question. i could imagine that you can have a feature completion on a split that an app doesn't own, but tbh, i have no idea.

jmileham:
One comment on wording

jmileham (commented):
New GitHub app caught me out - does this submit my line comment?

app/models/split.rb
jmileham: An app def can have a feature completion on a split owned by another app. Remote kills are also per app and can be against a split owned by any app.

app/views/admin/splits/_feature_completions.html.erb
jmileham: Yup it does.

app/views/admin/splits/_remote_kills.html.erb
jmileham: Yup that was so PG could sort it for us in semver order.

app/models/split.rb
samandmoore: right! i'm limiting this too much. good call. i'm gonna make it a regular association so we properly show all the completions and remote kills across all apps for the split

Gemfile
```diff
@@ -78,8 +80,9 @@ end
 
 group :development do
   # Access an IRB console on exception pages or by using <%= console %> in views
-  gem 'web-console', '~> 2.0'
+  gem 'web-console', '>= 3.3.0'
   gem 'travis', '~> 1.8.0'
+  gem 'listen', '~> 3.1.5'
 end
```
samandmoore, line 86: i was having all sorts of issues running the app locally and i think it's because when this app was upgraded to rails 5.2, i don't think we ran `rails app:update`. so i just ran that now and committed the parts of the diff that we want.

samandmoore: there is still an issue with hot reloading when constants change, but we can figure that out next.

app/views/admin/splits/_feature_completions.html.erb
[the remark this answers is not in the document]
samandmoore: updated!

samandmoore:
bump! 

i ended up having to sneak some rails 5.2 stuff in here because the app was not working properly locally without those changes. there's still some funkiness with auto-loading, but we can figure that out separately

app/models/app_feature_completion.rb
```diff
@@ -4,7 +4,7 @@ class AppFeatureCompletion < ActiveRecord::Base
 
   attribute :version, :app_version
 
-  validates :app, :feature_gate, :version, presence: true
+  validates :app, :version, presence: true
```
samandmoore, line 7: rails now does this automatically for belongs_to associations by default

app/models/app_remote_kill.rb
```diff
@@ -5,14 +5,16 @@ class AppRemoteKill < ActiveRecord::Base
   attribute :first_bad_version, :app_version
   attribute :fixed_version, :app_version
 
-  validates :app, :split, :reason, :override_to, :first_bad_version, presence: true
+  validates :app, :reason, :override_to, :first_bad_version, presence: true
```
samandmoore, line 8: same here, rails does this for us

config/application.rb
```diff
@@ -29,6 +29,8 @@ class Application < Rails::Application
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     # config.i18n.default_locale = :de
 
+    config.load_defaults 5.0
```
samandmoore, line 32: hm, i worried about this. lemme take a look at what this would have been doing if we weren't calling it but we were using 5.2

samandmoore: okay! based on this https://guides.rubyonrails.org/configuring.html#results-of-config-load-defaults and conversation with the folks that have done BMT rails upgrades, we feel good about this change to load_defaults 5.2

samandmoore:
bump @aburgel 

i'm just trying to determine if we can remove some of the environments/ files

samandmoore:
> i'm just trying to determine if we can remove some of the environments/ files

determined! and deleted.

aburgel:
domainlgtm!

looks like travis is choking on the ruby 2.4 build. i'd say it's time to retire that version.

nanda-prbot (approved):
Approved! :taco: :100: :dart:

PR #148, opened 2021-01-15
new versions of app_visitor_configs and app_identifier_visitor_configs

samandmoore:
### Summary

new versions of app_visitor_configs and app_identifier_visitor_configs

* introduce a new, deserialization friendly split registry format (array
  of splits is nicer to work with than a map of split name => registry)
* remove unsynced and context from assignments response (no longer used
  by clients)

these new endpoints will be used in by our newer versions of the mobile clients.

### Other Information

- [x] ~should this be v3?~ went with v4 🎉 
- [x] ~should we also introduce vNext endpoints using this split registry structure for _all of the other_ split registry endpoints that are still actively in-use? i think that's [`/v3/builds/:timestamp/split_registry`](https://github.com/Betterment/test_track_rails_client/blob/81269da02e4290f31a75f3fef569ab1cbbbddb50/app/models/test_track/remote/split_registry.rb#L6) which powers the rails client and the js client~ went with not yet. we should have a v4 of the builds/.../split_registry endpoint but we don't need it yet.

/no-platform
/cc @mseijas @ClaireDavis @btrautmann


app/controllers/api/v2/app_identifier_visitor_configs_controller.rb
```diff
@@ -0,0 +1,26 @@
+class Api::V2::AppIdentifierVisitorConfigsController < UnauthenticatedApiController
```
samandmoore, line 1: i copied these controllers verbatim from v1

app/controllers/api/v2/app_visitor_configs_controller.rb
```diff
@@ -0,0 +1,26 @@
+class Api::V2::AppVisitorConfigsController < UnauthenticatedApiController
```
samandmoore, line 1: i copied these controllers verbatim from v1

app/views/api/v2/app_visitor_configs/_show.json.jbuilder
```diff
@@ -0,0 +1,10 @@
+json.splits active_splits do |split|
+  json.name split.name
+  json.weights split.registry
+  json.feature_gate split.feature_gate?
+end
```
samandmoore, line 5: this is part of the meaningful changes in the PR: now this is an array instead of a map

app/views/api/v2/app_visitor_configs/_show.json.jbuilder
```diff
@@ -0,0 +1,10 @@
...
+
+json.visitor do
+  json.id visitor_id
+  json.assignments assignments, :split_name, :variant
```
samandmoore, line 9: and we dropped the context and unsynced attributes here since they're related to syncing unsynced assignments to analyticsland on fetch of the visitor assignments which we're not going to be doing anymore in clients

config/routes.rb
```diff
@@ -58,6 +58,21 @@
...
+              resource :config, only: :show, controller: 'app_visitor_configs'
+            end
+            resources :identifier_types, only: [], param: :name do
+              resources :identifiers, only: [], param: :value do
+                resource :visitor_config, only: :show, controller: 'app_identifier_visitor_configs'
+              end
+            end
+          end
+        end
+      end
```
samandmoore, line 74: copy paste from v1

spec/controllers/api/v2/app_identifier_visitor_configs_controller_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        version_number: "1.0",
+        build_timestamp: "2019-04-16T14:35:30Z",
+        identifier_type_name: identifier.identifier_type.name,
+        identifier_value: identifier.value
+      }
+
+      expect(response).to have_http_status :ok
+      expect(response_json['splits'][0]['name']).to eq('blab_enabled')
+      expect(response_json['splits'][0]['weights']).to eq('false' => 100, 'true' => 0)
+      expect(response_json['splits'][0]['feature_gate']).to eq(true)
```
samandmoore, line 32: everything except this chunk is copy paste

spec/controllers/api/v2/app_identifier_visitor_configs_controller_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        version_number: "1.0",
+        build_timestamp: "2019-04-16T14:35:30Z",
+        identifier_type_name: identifier.identifier_type.name,
+        identifier_value: identifier.value
+      }
+
+      expect(response).to have_http_status :ok
+      expect(response_json['splits'][0]['name']).to eq('blab_enabled')
+      expect(response_json['splits'][0]['weights']).to eq('false' => 50, 'true' => 50)
+      expect(response_json['splits'][0]['feature_gate']).to eq(true)
```
samandmoore, line 64: everything except this chunk is copy paste

spec/controllers/api/v2/app_visitor_configs_controller_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+        app_name: app.name,
+        version_number: "1.0",
+        build_timestamp: "2019-04-16T14:35:30Z",
+        visitor_id: visitor.id
+      }
+
+      expect(response).to have_http_status :ok
+      expect(response_json['splits'][0]['name']).to eq('blab_enabled')
+      expect(response_json['splits'][0]['weights']).to eq('false' => 100, 'true' => 0)
+      expect(response_json['splits'][0]['feature_gate']).to eq(true)
```
samandmoore, line 29: everything except this chunk is copy paste

spec/controllers/api/v2/app_visitor_configs_controller_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+        version_number: "1.0",
+        build_timestamp: "2019-04-16T14:35:30Z",
+        visitor_id: visitor.id
+      }
+
+      expect(response).to have_http_status :ok
+
+      expect(response_json['splits'][0]['name']).to eq('blab_enabled')
+      expect(response_json['splits'][0]['weights']).to eq('false' => 50, 'true' => 50)
+      expect(response_json['splits'][0]['feature_gate']).to eq(true)
```
samandmoore, line 61: everything except this chunk is copy paste

spec/controllers/api/v2/app_identifier_visitor_configs_controller_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        app_name: app.name,
+        version_number: "1.0",
+        build_timestamp: "2019-04-16T14:35:30Z",
+        identifier_type_name: identifier.identifier_type.name,
+        identifier_value: identifier.value
+      }
+
+      expect(response).to have_http_status :ok
+      expect(response_json['splits'][0]['name']).to eq('blab_enabled')
+      expect(response_json['splits'][0]['weights']).to eq('false' => 100, 'true' => 0)
```
samandmoore, line 31: i just realized that we also wanna make one more change here to convert this into an array of variant objects instead of a map!

samandmoore: done!

app/views/api/v2/app_visitor_configs/_show.json.jbuilder
```diff
@@ -0,0 +1,15 @@
...
+    json.weight weight
+  end
+
+  json.feature_gate split.feature_gate?
+end
+
+json.visitor do
+  json.id visitor_id
+  json.assignments assignments, :split_name, :variant
+end
```
samandmoore, line 15: new format example:

```json
{
  "splits":[
    {
      "name":"blab_enabled",
      "variants":[
        {
          "name":"false",
          "weight":100
        },
        {
          "name":"true",
          "weight":0
        }
      ],
      "feature_gate":true
    }
  ],
  "visitor":{
    "id":"4ee64572-b00a-494f-a191-f4c40d6a390c",
    "assignments":[
      
    ]
  }
}
```

jmileham:
I like introducing a new API version (v3) if we're adding a new shape. Will be easier to shim into deserialization code in clients if all the new stuff is named V3. Also if we have high confidence about the shape we'll need for other clients, we could def cut the endpoints now so it's a smaller lift to upgrade them later (or add new clients in new languages that get to benefit from the better shape off the bat).

app/controllers/api/v2/app_identifier_visitor_configs_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+  include CorsSupport
+
+  def show # rubocop:disable Metrics/AbcSize
+    build_path = AppVersionBuildPath.new(build_params)
+    if build_path.valid?
+      app_build = AppVersionBuildPath.new(build_params).app_build
```
jmileham, line 7: why not `build_path.app_build`?

spec/controllers/api/v2/app_identifier_visitor_configs_controller_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    let(:identifier) { FactoryBot.create(:identifier, visitor: visitor) }
+
+    it "has knocked-out weightings and it doesn't include a non-force assignment for a feature gate that isn't feature complete" do
+      FactoryBot.create(
+        :assignment,
+        visitor: visitor,
+        split: feature_gate,
+        variant: "true",
+        context: "the_context",
+        mixpanel_result: "success"
```
jmileham, line 18: we can probably elide context and mixpanel result in factory config now that they're not returned?

jmileham:
<< domain tafn

app/controllers/api/v2/app_identifier_visitor_configs_controller.rb
samandmoore: great question! i dunno. this was a copy-paste job, but i went in and fixed it in all 4 places.

samandmoore:
> I like introducing a new API version (v3) if we're adding a new shape. Will be easier to shim into deserialization code in clients if all the new stuff is named V3. Also if we have high confidence about the shape we'll need for other clients, we could def cut the endpoints now so it's a smaller lift to upgrade them later (or add new clients in new languages that get to benefit from the better shape off the bat).

we already have 1 thing in the api/v3 namespace which is a split_registry endpoint nested under builds. it’s what’s used by the current version of the rails client. and it uses a split registry format that’s newer than v2 but not the same as what i’m doing for these 2 endpoints. so does that mean we should go to v4?

i’d expect that we’d wanna have a v4 endpoint for `/builds/:timestamp/split_registry` for the rails client that uses this new format eventually, but i was thinking of not adding that until we wanna use it. there's no urgency to change the serialization format for the rails client since it's happily consuming the map-like approach

samandmoore:
bump er ooooo

jmileham:
Domain lgtm!

nanda-prbot (approved):
Approved! :sparkles: :avocado: :hamburger:

PR #149, opened 2021-01-19
More v4 endpoint additions

samandmoore:
### Summary

in #148 we discussed creating a v4 version of the /builds/:timestamp/split_registry endpoint that is used by the rails client and js client (transitively).

this PR adds that endpoint for future consumption. this is nice in that it allows all the clients to be consuming the same version of the split registry schema (v4) whenever they're updated.

this PR also introduces the `experience_sampling_weight` to the v4 endpoints.

**NOTE: i moved the `experience_sampling_weight` loading behavior from the `SplitRegistry` model to the `Rails.configuration` in `application.rb`. This felt like a better way to share the behavior between the `SplitRegistry`-based and the non-`SplitRegistry`-based  ways of getting at the list of applicable splits for a client (via app+version+timestamp or via build timestamp) cc @jmileham @aburgel the main result of this move is that the tests covering the tiny bit of behavior for validating the experience sampling weight are gone**

/no-platform

### DIFF
--- a/spec/models/split_registry_spec.rb
+++ b/spec/models/split_registry_spec.rb
@@ -49,39 +49,4 @@ RSpec.describe SplitRegistry do
       expect(described_class.new(as_of: 1.day.ago).splits.all).to include(split)
     end
   end
-
-  describe "#experience_sampling_weight" do
-    it "memoizes the env var fetch" do
-      allow(ENV).to receive(:fetch).and_call_original
-
-      subject.experience_sampling_weight
-      subject.experience_sampling_weight
-
-      expect(ENV).to have_received(:fetch).with('EXPERIENCE_SAMPLING_WEIGHT', any_args).exactly(:once)
-    end
-
-    it "returns 1 with no env var" do
-      with_env EXPERIENCE_SAMPLING_WEIGHT: nil do
-        expect(subject.experience_sampling_weight).to eq 1
-      end
-    end
-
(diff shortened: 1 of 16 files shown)


spec/models/split_registry_spec.rb
```diff
@@ -49,39 +49,4 @@
...
-
-      subject.experience_sampling_weight
-      subject.experience_sampling_weight
-
-      expect(ENV).to have_received(:fetch).with('EXPERIENCE_SAMPLING_WEIGHT', any_args).exactly(:once)
-    end
-
-    it "returns 1 with no env var" do
-      with_env EXPERIENCE_SAMPLING_WEIGHT: nil do
-        expect(subject.experience_sampling_weight).to eq 1
```
smudge, line 65: Now that this is gone, is there code somewhere that tests the default sampling weight without stubbing?

smudge: (I get that the tests for the `raise` check are also gone, but I'm just wondering if it's still worth exercising the full codepath at least once in a spec, to avoid surprise breakages in prod.)

smudge:
<< domain TAFN

samandmoore:
bump!

smudge:
domain LGTM!

nanda-prbot (approved):
Approved! :burrito: :ok_hand: :+1:

PR #150, opened 2021-01-28
Add V4 Identifier POST endpoint, remove unused V4 GET endpoint

samandmoore:
### Summary

* TT client side log-ins are done with a combination of two API calls: `POST /v1/identifier` followed by a `GET` to a v4 app version build path visitor configs endpoint. We'd like to replace this POST + GET chain with a single endpoint that performs the combined behavior of both endpoints.

* We'd like to remove the `GET /v4/apps/:app_id/versions/:version_id/builds/:timestamp/identifier_types/:identifier_type/identifiers/:value/visitor_config` endpoint. This endpoint is not used and never has been.

The new endpoint will look like this:

`POST /v4/apps/:app_id/versions/:version_id/builds/:timestamp/identifier`
`{ identifier_type: 'broker_dealer_user_id', value: '123', visitor_id: 'current client-side visitor id' }`

### Other Information

The v4 app version build path identifier visitor configs endpoint was recently added, but never used, so we feel good deleting it. 

/no-platform


app/controllers/api/v4/identifiers_controller.rb
```diff
@@ -0,0 +1,32 @@
+class Api::V4::IdentifiersController < UnauthenticatedApiController
```
samandmoore, line 1: i don't love this name, but the other names i considered didn't feel better:
* `AppIdentifiersController`
* `AppVersionBuildPathIdentifiersController`
* `AppVersionBuildPathIdentifierVisitorConfigsController`

🤮 

The main thing i don't love about this name is that it doesn't indicate that this controller is nestled within "app version build path" stuff. In other other controllers where we do this, we call the controller `app*` but `AppIdentifiers` feels like a confusing name, not a clarifying one. Given that, i decided to stick with just Identifiers.

app/controllers/api/v4/identifiers_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+        @visitor_id = visitor.id
+        @assignments = visitor.assignments_for(app_build).includes(:split).order(:updated_at)
+        @experience_sampling_weight = Rails.configuration.experience_sampling_weight
+      else
+        render_errors claim
+      end
+    else
+      render_errors build_path
+    end
+  end
```
samandmoore, line 21: this is basically the max logic i'm okay with putting in a controller. creating a new model for this just feels unnecessary to me.

spec/controllers/api/v4/identifiers_controller_spec.rb
```diff
@@ -1,16 +1,15 @@
 require 'rails_helper'
 
-RSpec.describe Api::V4::AppIdentifierVisitorConfigsController do
```
samandmoore, line 3: these tests are basically a mash-up of the v1 identifier controller spec and the v4 app visitor config spec

app/controllers/api/v4/identifiers_controller.rb
jmileham: I think you should go with the App prefix, otherwise you've got two different version numbers of the same controller name that have different semantics, are for different client types, and are both the latest version of their API endpoint for their use case.

app/controllers/api/v4/identifiers_controller.rb
jmileham: I think there could be a case for it if we changed the JSON structural representation in a future version of this endpoint, then we'd have exactly the same controller code copypasta'd. Does that move you?

jmileham:
<< domain
<< platform
tafn

app/controllers/api/v4/identifiers_controller.rb
samandmoore: it doesn't really move me 😄 

my vote is to refactor this into a model as a part of that future work to avoid the copypasta because at least then we'd know that we're actually reusing the same logic in more than one place. but really i'm just trying to avoid having to create a new model that figures out how to coordinate between these two different models right now 😛

samandmoore: ah, F-it. i decided to just dive into this.

app/models/app_identifier_claim.rb
```diff
@@ -0,0 +1,50 @@
+class AppIdentifierClaim
```
samandmoore, line 1: added this model that composes the other two models.

spec/models/app_identifier_claim_spec.rb
```diff
@@ -0,0 +1,98 @@
+require 'rails_helper'
+
+RSpec.describe AppIdentifierClaim do
```
samandmoore, line 3: added minimal tests for this model:
* showing the validations carry through from the underlying models
* showing that save depends on the underlying models
* showing that the post-save accessors are derived from the other models

spec/models/app_identifier_claim_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+      identifier = instance_double(Identifier, visitor: visitor)
+      identifier_claim = instance_double(IdentifierClaim, valid?: true, save!: true, identifier: identifier)
+      allow(IdentifierClaim).to receive(:new).and_return identifier_claim
+
+      claim = described_class.new create_params
+      expect(claim.save).to eq(true)
+
+      expect(claim.app_build).to eq(app_build)
+    end
+  end
```
samandmoore, line 119: the stubbing in some of these tests is a lil beefy, but the alternative is to use factory bot to set up the db models so i at least wanted to try this out and see if others felt okay about it.

samandmoore:
bump @jmileham thoughts?

jmileham:
This looks awesome.

Domain lgtm platform lgtm!

nanda-prbot (approved):
Approved! :fireworks: :dancers: :ok_woman:

PR #155, opened 2021-07-12
Replace Travis with GH Actions

smudge:
### Summary

Travis switched from .org to .com and our config didn't transfer over. Rather than trying to fix that, I decided to take a stab at switching this to GitHub Actions!

Mostly aiming to get tests running & passing here. It might take a few tries!

/no-platform


.github/workflows/ci.yml
```diff
@@ -0,0 +1,41 @@
...
+      postgres:
+        image: postgres:11.0
+        env:
+          POSTGRES_USER: runner
+          POSTGRES_PASSWORD: ""
+          POSTGRES_DB: test_track_test
+        ports:
+          - 5432:5432
+        options: >-
+          --mount type=tmpfs,destination=/var/lib/postgresql/data
```
smudge, line 26: mounting the db in a tmpfs (in memory) speeds things up!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,41 @@
...
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
```
smudge, line 30: The health check config ensures that the container setup waits for pg to start running.

samandmoore:
<<domainlgtm

any reason you're using ruby 2.6.1 instead of 2.6.2?

nanda-prbot (approved):
Approved! :avocado: :zap: :golf:

smudge:
> any reason you're using ruby 2.6.1 instead of 2.6.2?

Answered offline! I think the next step here is to introduce a real test matrix for Ruby versions, but I'm holding off for a future PR.

PR #168, opened 2022-07-20
Build against Ruby 2.7

smudge:
### Summary

This moves the repo onto Ruby 2.7.5 by default. I tried to make as few gem upgrades as I could in order to get onto 2.7, but it did require a move onto a newer capybara/selenium-webdriver, a newer rubocop, and a newer mini_racer (with libv8-node). I also removed poltergeist and got tests working with chromdriver in CI.

### Other Information

Not much else! Stay tuned for some more updates after this!

/no-platform


agirlnamedsophia (approved):
<< domainlgtm

nanda-prbot (approved):
Approved! :hamburger: :star: :clap:

PR #178, opened 2023-02-20
UI Updates (Filter by App, Search by Split, Sort by Date)

jesseproudman:
### Summary
Updates the UI to provide a few helpful features:

- Allows you to search by a split name from the box in the header
- Allows you to filter by App
- Allows you to change sorting to be by name or date

### DIFF
--- a/app/controllers/admin/split_searches_controller.rb
+++ b/app/controllers/admin/split_searches_controller.rb
@@ -1,0 +1,10 @@
+class Admin::SplitSearchesController < AuthenticatedAdminController
+  def create
+    split = Split.find_by(name: params[:name])
+
+    return redirect_to admin_split_path(split) if split
+
+    flash[:error] = "No split with that name could be found."
+    redirect_to admin_root_path
+  end
+end
--- a/app/controllers/admin/split_searches_controller.rb
+++ b/app/controllers/admin/split_searches_controller.rb
@@ -4,7 +4,7 @@ class Admin::SplitSearchesController < AuthenticatedAdminController
 
     return redirect_to admin_split_path(split) if split
 
-    flash[:error] = "No split with that name could be found."
+    flash[:error] = t('split_searches.no_results')
     redirect_to admin_root_path
   end
 end
--- a/app/controllers/admin/splits_controller.rb
+++ b/app/controllers/admin/splits_controller.rb
@@ -1,9 +1,26 @@
 class Admin::SplitsController < AuthenticatedAdminController
   def index
-    @splits = Split.active.order(:name)
+    if params[:app].present?
+      @app = App.find_by(name: params[:app])
+      return redirect_to admin_root_path unless @app
+
+      @splits = @app.splits
+    else
+      @splits = Split.all
+    end
+
+    @splits = @splits.active.order(sort_field)
   end
 
   def show
     @split = Split.find(params[:id])
+  end
+
+  private
+
+  def sort_field
+    return :created_at if params[:sort] == 'date'
+
+    :name
   end
 end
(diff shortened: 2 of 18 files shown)


app/controllers/admin/split_searches_controller.rb
```diff
@@ -0,0 +1,10 @@
+class Admin::SplitSearchesController < AuthenticatedAdminController
+  def create
+    split = Split.find_by(name: params[:name])
```
smudge, line 3: So, just confirming that this is intentional:
- this doesn't filter down to active splits
- if you've filtered the index down to a specific app, the search will ignore that filter

I think these are very reasonable choices. I don't think there's any reason you should'nt be able to get back to a fully retired split, as long as it's very clear that it's retired in the UI.

app/views/admin/splits/index.html.erb
```diff
@@ -1,19 +1,36 @@
...
+      for <%= @app.name %>
+    <% end %>
+  </h1>
+  <% if App.by_name.any? %>
+    <select onchange="if (this.value) window.location.href=this.value" data-testId="appSelector">
+      <option value="">Filter by App</option>
+      <% App.by_name.each do |app| %>
+        <option value="<%= admin_root_path(app: app.name) %>"><%= app.name %></option>
+      <% end %>
+    </select>
```
HipsterBrown, line 14: Can this use a regular form GET submission instead of this inline change handler? Redirecting based on selecting an item has unexpected accessibility issues. With a form submission, we don't need to provide the whole path as the value for each option either, just set the name and id on the select to `app` and the value of each option to `app.name`. 

The select should have an associated label as well rather than using an empty option as the identifier.

app/controllers/admin/splits_controller.rb
```diff
@@ -1,9 +1,26 @@
...
   end
 
   def show
     @split = Split.find(params[:id])
   end
+
+  private
+
+  def sort_field
+    return :created_at if params[:sort] == 'date'
```
smudge, line 22: thoughts on a configurable sort order as well? (asc vs desc)

app/controllers/admin/split_searches_controller.rb
jesseproudman: Yes, the goal was to be able to find retired splits.

And yes, it was meant to ignore the app.

app/controllers/admin/splits_controller.rb
jesseproudman: Didn't feel it was worth the hassle tbh - main goal here is to be able to see the oldest splits to know what we we can go after attacking. WDYT?

app/views/admin/splits/index.html.erb
```diff
@@ -1,19 +1,36 @@
...
-    </thead>
-    <tbody>
-      <%= render @splits %>
-    </tbody>
-  </table>
+  <% if @splits.any? %>
+    <table class="BasicTable">
+      <thead>
+        <tr>
+          <th><a href="?sort=name">Name</a></th>
```
smudge, line 23:
```suggestion
          <th><%= link_to('Name', path_for(sort: 'name')) %></th>
```

app/controllers/admin/splits_controller.rb
smudge: Yeah that seems reasonable to me. There's some UI component reworking I'd love to do so that we'd get more of these features for free anyways. Happy to leave that for later.

app/views/admin/splits/index.html.erb
jesseproudman: Screen Shot 2023-02-28 at 3 11 23 PM

jesseproudman:
Just going to use this query instead:

https://redash.betterment.io/queries/10871

s/o @phantomphildius for the link.

PR #181, opened 2023-02-20
Add Github search link if GITHUB_ORGANIZATION ENV variable is set

jesseproudman:
### Summary

When removing splits, it's important to verify if they're no longer in use across the code base. Right now, that process is manual and tedious and challenging for others to replicate. This PR brings a link to the UI (split table and split detail) that provides a quick reference to Github to validate that splits are no longer active.

Screen Shot 2023-02-20 at 3 33 47 PM

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -6,4 +6,14 @@ module ApplicationHelper
   def percentage(ratio)
     "#{(ratio * 100).round}%"
   end
+
+  def github_search_enabled?
+    ENV['GITHUB_ORGANIZATION'].present?
+  end
+
+  def github_search_url(split)
+    return nil unless github_search_enabled?
+
+    "https://github.com/search?q=org%3A#{ENV['GITHUB_ORGANIZATION']}+#{split.name}&type=code"
+  end
 end
--- a/spec/system/admin_split_details_create_spec.rb
+++ b/spec/system/admin_split_details_create_spec.rb
@@ -17,6 +17,20 @@ RSpec.describe 'admin can add details to a split' do
     login
   end
 
+  context 'if GITHUB_ORGANIZATION ENV is set' do
+    before do
+      ENV['GITHUB_ORGANIZATION'] = 'test'
+    end
+
+    it 'displays Github search link' do
+      split_page.load split_id: split.id
+
+      expect(split_page).to be_loaded
+      expect(split_page).to have_text("Github")
+      expect(split_page).to have_text("Search for uses")
+    end
+  end
+
   it 'allows admins to add details to a split' do
     split_page.load split_id: split.id
     expect(split_page).to be_loaded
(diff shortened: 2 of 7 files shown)


app/helpers/application_helper.rb
```diff
@@ -6,4 +6,14 @@ def identifier_types
...
+
+  def github_search_enabled?
+    ENV['GITHUB_ORGANIZATION'].present?
+  end
+
+  def github_search_url(split)
+    return nil unless github_search_enabled?
+
+    "https://github.com/search?q=org%3A#{ENV['GITHUB_ORGANIZATION']}+#{split.name}&type=code"
+  end
```
samandmoore, line 18: I think we should lift some of this code out into a configuration model. It's a bit wonky to have this "feature is enabled" code exist in a view helper IMO. lemme sketch out a way to do that...

app/helpers/application_helper.rb
```diff
@@ -6,4 +6,14 @@ def identifier_types
   def percentage(ratio)
     "#{(ratio * 100).round}%"
   end
+
+  def github_search_enabled?
+    ENV['GITHUB_ORGANIZATION'].present?
+  end
+
+  def github_search_url(split)
+    return nil unless github_search_enabled?
```
smudge, line 15: One might argue that this could `raise`, since the existence of `github_search_enabled?` implies that you shouldn't call this method unless that returns true.

smudge: Alternatively, this could be a nullable facet of the split model:

`@split.github_search_url`

And the null/presence could be checked via a predicate method -- `@split.github_search_url?`

app/helpers/application_helper.rb
```diff
@@ -6,4 +6,14 @@ def identifier_types
   def percentage(ratio)
     "#{(ratio * 100).round}%"
   end
+
+  def github_search_enabled?
+    ENV['GITHUB_ORGANIZATION'].present?
```
samandmoore, line 11: can you add a note to the README about this feature? 

it would be useful to document this as a configurable env variable that enables this functionality and what it's useful for.

app/helpers/application_helper.rb
samandmoore: okay, so here's my refactoring 81a5c539472a8bb593749a2a9e903c94b73039a6

spec/system/admin_split_details_create_spec.rb
```diff
@@ -17,6 +17,20 @@
     login
   end
 
+  context 'if GITHUB_ORGANIZATION ENV is set' do
+    before do
+      ENV['GITHUB_ORGANIZATION'] = 'test'
```
samandmoore, line 22: we should use our `with_env` helper here so that we unset this after the test case runs. it looks like `with_env('THING' => 'VALUE') do ... end`

jesseproudman:
LMK how I can help land this.

samandmoore:
I left a bunch of feedback. I'm happy to take this over and integrate that. I think those are the only things I see as blockers to shipping this.

PR #182, opened 2023-02-21
feat: upgrade Rails to v6.0.6.1

HipsterBrown:
### Summary

To support new feature development and replacing older, vendored dependencies, this PR upgrades Rails from 5.2.4 to 6.0.6.1, along with associated dependencies through `bundle update`. One of these upgrades includes `betterlint`, so most of the changes are automatic fixes to Rubocop rules. 

### Other Information


Gemfile
```diff
@@ -25,7 +25,7 @@ gem 'nokogiri'
 
 gem 'responders'
 
-gem 'rack-timeout'
+gem 'rack-timeout', require: "rack/timeout/base"
```
HipsterBrown, line 28: In order to configure the service_timeout, this change is required for manual middleware initialization AFAICT: https://github.com/zombocom/rack-timeout#rails-apps-manually

I was getting an error saying `service_timeout` was not an available method otherwise.

app/assets/config/manifest.js
```diff
@@ -0,0 +1,3 @@
+//= link_tree ../images
+//= link_directory ../javascripts .js
+//= link_directory ../stylesheets .css
```
HipsterBrown, line 3: `manifest.js` is required in newer versions of Sprockets

samandmoore:
just wanted to check in here at a high-level... did you follow along with the upgrade instructions here when you went through this? https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-2-to-rails-6-0

with rails upgrades there's usually a bunch of small to medium gotchas that are captured in those guides. in this case, going from 5.2 to 6.0 there were a buncha things to consider.

config/application.rb
```diff
@@ -29,7 +29,7 @@ class Application < Rails::Application
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     # config.i18n.default_locale = :de
 
-    config.load_defaults 5.2
+    config.load_defaults 6.0
```
samandmoore, line 32: looks like safe changes re: what this app does https://guides.rubyonrails.org/configuring.html#default-values-for-target-version-6-0

cc @smudge when you have a chance since i know you've got a deep history of thoughts on 5.2 -> 6.0 -- if you remember anything

smudge: Taking a look...

I also agree on the webpacker front. 🙂

config/initializers/rack_timeout.rb
```diff
@@ -1,11 +1,15 @@
-Rack::Timeout.service_timeout = 30
+# Rack::Timeout.service_timeout = 30
+
+# insert middleware wherever you want in the stack, optionally pass
+# initialization arguments, or use environment variables
+Rails.application.config.middleware.insert_before Rack::Runtime, Rack::Timeout, service_timeout: 30
```
smudge, line 5: I don't think this is something we've had to do elsewhere, so I'd like to get to the bottom of why we had to do it here -- and maybe then we can also undo [this change](https://github.com/Betterment/test_track/pull/182/files#r1113689994) as well.

app/controllers/shared_secret_authenticated_api_controller.rb
```diff
@@ -16,6 +16,6 @@ def authenticate
   end
 
   def shared_secret
-    ENV[SHARED_SECRET_ENV_VAR]
+    ENV.fetch(SHARED_SECRET_ENV_VAR, nil)
```
smudge, line 19: Just confirming, the `nil` return is fairly safe here because up in `authenticate` we're `raise`ing if it's nil.

app/inputs/collection_select_input.rb
```diff
@@ -116,7 +116,7 @@ def include_blank_as_string
 
   def prompt_as_string
     if options[:prompt] == true
-      I18n.translate('helpers.select.prompt', default: 'Please select')
+      I18n.t('helpers.select.prompt', default: 'Please select')
```
smudge, line 119: minor, but adding `!` will raise on missing translations. I think this is our preferred behavior:
 
```suggestion
      I18n.t!('helpers.select.prompt', default: 'Please select')
```

smudge: (aligning this entire app to that behavior might just be worth separate PRs)

app/models/app_feature_completion.rb
```diff
@@ -4,7 +4,7 @@ class AppFeatureCompletion < ActiveRecord::Base
 
   attribute :version, :app_version
 
-  validates :app, :version, presence: true
+  validates :version, presence: true
```
smudge, line 7: Is this because `belongs_to :app` should already enforce presence?

app/models/split.rb
```diff
@@ -201,7 +201,7 @@ def variants_must_be_snake_case
   end
 
   def registry_must_have_winning_variant_if_decided
-    errors.add(:registry, "must have a winning variant if decided") if decided_at.present? && registry.values.none? { |v| v == 100 }
+    errors.add(:registry, "must have a winning variant if decided") if decided_at.present? && registry.values.none?(100)
```
smudge, line 204: Huh, not sure I knew about this shorthand.

config/application.rb
smudge: I don't think any of these 6.0 behaviors really apply to this application, or would be things that would make it by without a test failing somewhere.

config/environments/test.rb
```diff
@@ -5,7 +5,7 @@
   # test suite. You never need to work with it otherwise. Remember that
   # your test database is "scratch space" for the test suite and is wiped
   # and recreated between test runs. Don't rely on the data there!
-  config.cache_classes = true
+  config.cache_classes = false
```
smudge, line 8: Here's what we like to stick in `test.rb`:

```ruby
  config.cache_classes = ActiveRecord::Type::Boolean.new.cast(ENV.fetch('CACHE_CLASSES', '1'))
  config.eager_load = ActiveRecord::Type::Boolean.new.cast(ENV.fetch('EAGER_LOAD', '0'))
  config.assets.compile = ActiveRecord::Type::Boolean.new.cast(ENV.fetch('COMPILE_ASSETS', '1'))
```

This gives us finer-grained controls for running tests locally (where we don't want to eager load all classes) vs in CI (where we _do_ want to fully eager load everything, and also precompile and test with production-like asset behaviors).

smudge: Similarly, here's what we could probably stick in `development.rb`:

```ruby
  config.cache_classes = ActiveRecord::Type::Boolean.new.cast(ENV.fetch('CACHE_CLASSES', '0'))
  config.eager_load = ActiveRecord::Type::Boolean.new.cast(ENV.fetch('EAGER_LOAD', '0'))
  config.assets.compile = ActiveRecord::Type::Boolean.new.cast(ENV.fetch('COMPILE_ASSETS', '1'))
```

smudge: I could see a case for having `'0'` for `CACHE_CLASSES` in test.rb, if that's the new default.

config/initializers/airbrake.rb
```diff
@@ -9,7 +9,7 @@
       config.host = begin
         host_uri = URI::parse(ENV['AIRBRAKE_HOST'])
         port = ENV['AIRBRAKE_PORT'].presence&.to_i || host_uri.port
-        uri_builder = port == 443 ? URI::HTTPS : URI::HTTPS
+        uri_builder = URI::HTTPS
```
smudge, line 12: I wonder if this was a typo, and we intended to have some `URI::HTTP` support at some point?

spec/controllers/api/v1/app_identifier_visitor_configs_controller_spec.rb
```diff
@@ -2,14 +2,14 @@
...
-    let(:identifier) { FactoryBot.create(:identifier, visitor: visitor) }
+    let(:app) { create(:app) }
+    let(:identifier_type) { create(:identifier_type, name: "clown_id") }
+    let(:feature_gate) { create(:feature_gate, name: "blab_enabled", registry: { false: 50, true: 50 }) }
+    let(:visitor) { create(:visitor) }
+    let(:identifier) { create(:identifier, visitor: visitor) }
 
     it "has knocked-out weightings and it doesn't include a non-force assignment for a feature gate that isn't feature complete" do
-      FactoryBot.create(
+      create(
```
smudge, line 12: I think we typically don't mix in / rely on FactoryBot's spec helper mixin -- instead going with the explicit `FactoryBot.create(...)` calls (etc).

Thoughts? Was this a linter autocorrect that we can revert?

smudge (changes requested):
TAFN -- most of this seems totally fine, but I zoomed in on a few things that jumped out at me. It might also make sense to split the rubocop autocorrections out into a series of PRs, so that if something goes wrong, we can more quickly isolate the specific issue and rollback.

app/inputs/collection_select_input.rb
HipsterBrown: This was an automated fix from rubocop. I'm happy to split out the rubocop fixes into a separate PR as long as CI doesn't block on the linter.

app/models/app_feature_completion.rb
HipsterBrown: Yes, it is caught as a redundant validation for relationships.

config/initializers/airbrake.rb
HipsterBrown: I can change it back to set `URI:HTTP` as the else case if we'd like.

spec/controllers/api/v1/app_identifier_visitor_configs_controller_spec.rb
HipsterBrown: This was an automated linter fix. I can revert the fix and turn off this rule if that's our standard.

HipsterBrown:
@smudge @samandmoore I've upgraded betterlint to the latest release (v1.4.0) to resolve the FactoryBot SyntaxMethods rule changes. Thanks for your review!

smudge (approved):
domain LGTM && platform LGTM

PR #183, opened 2023-03-02
[Ruby 3] Test Track Upgrade

jesseproudman:
### Summary

Upgrades test_track to Ruby 3 support.


Gemfile
```diff
@@ -8,8 +8,6 @@ gem 'pg'
 gem 'sass-rails'
 # Use Uglifier as compressor for JavaScript assets
 gem 'uglifier', '>= 1.3.0'
-# Use CoffeeScript for .coffee assets and views
-gem 'coffee-rails', '~> 4.2.0'
```
jesseproudman, line 12: As far as I can tell, this wasn't being used and throws this warning:

```
DEPRECATION WARNING: Single arity template handlers are deprecated. Template handlers must
now accept two parameters, the view object and the source for the view object.
Change:
  >> Coffee::Rails::TemplateHandler.call(template)
To:
  >> Coffee::Rails::TemplateHandler.call(template, source)
 (called from <top (required)> at /Users/jesseproudman/Source/betterment/test_track/config/environment.rb:5)
```

Gemfile
```diff
@@ -41,8 +39,7 @@ gem 'omniauth-rails_csrf_protection'
 
 gem 'simple_form'
 
-gem 'paperclip', '~> 5.2'
-gem 'aws-sdk', '~> 2.3.0'
+gem 'kt-paperclip', '~> 7.1'
```
jesseproudman, line 42: Original paperclip gem is deprecated and isn't compatible with Ruby 3.

Gemfile
```diff
@@ -81,7 +78,6 @@ group :development do
   # Access an IRB console on exception pages or by using <%= console %> in views
   gem 'web-console', '>= 3.3.0'
   gem 'travis', '~> 1.8.0'
-  gem 'listen', '~> 3.1.5'
```
jesseproudman, line 84: Listen is not compatible with Ruby 3.

jesseproudman:
@smudge This look ready for eyes.

vendor/gems/ruby_spec_helpers/lib/ruby_spec_helpers/capybara_configuration.rb
```diff
@@ -33,11 +33,10 @@
   when :selenium_chrome_headless
     Capybara.register_driver :selenium_chrome_headless do |app|
       args = %w(headless disable-gpu disable-dev-shm-usage no-sandbox window-size=1280,1024)
-      options_key = Selenium::WebDriver::VERSION >= '4' ? :capabilities : :options
 
       Capybara::Selenium::Driver.new app,
         browser: :chrome,
-        options_key => Selenium::WebDriver::Chrome::Options.new(args: args),
+        options: Selenium::WebDriver::Chrome::Options.new(args: args),
```
jesseproudman, line 39: This was to fix this deprecation notice:

```
Selenium [DEPRECATION] [:capabilities] The :capabilities parameter for Selenium::WebDriver::Chrome::Driver is deprecated. Use :options argument with an instance of Selenium::WebDriver::Chrome::Driver instead.
```

But I don't think is actually the fix... Need to sort this one out.

Gemfile
```diff
@@ -41,8 +39,7 @@ gem 'omniauth-rails_csrf_protection'
 
 gem 'simple_form'
 
-gem 'paperclip', '~> 5.2'
-gem 'aws-sdk', '~> 2.3.0'
```
samandmoore, line 45: We don't need this anymore?

vendor/gems/ruby_spec_helpers/lib/ruby_spec_helpers/capybara_configuration.rb
samandmoore: It seems like it's working correctly in CI. Are there issues locally?

Gemfile
samandmoore: This should be fine for now. We should make a plan to migrate to active storage.

Gemfile
samandmoore: Yup. We never used coffee script in this app. So we should have removed it sooner.

Gemfile
smudge: I think we do, but only in production.

Gemfile
smudge: There is one attachment type in this app, and we'll want to make sure that the newer version of paperclip can still find the previously-uploaded files in production.

smudge (changes requested):
I'm pretty sure that the `aws-sdk` dependency is how paperclip connects to S3 in production. But the AWS SDK has gone through some iterations, and `aws-sdk-s3` may be all we actually need to include in order to not break the production behavior.

(In test/dev it uses the local filesystem, which is why we can seemingly remove the gem and have nothing fail.)

vendor/gems/ruby_spec_helpers/lib/ruby_spec_helpers/capybara_configuration.rb
jesseproudman: No, no issues locally... the fix just logically doesn't seem right to me but I was reading it at 2am.

Gemfile
jesseproudman: I mis-read the [`kt-paperclip` docs](https://betterment.slack.com/archives/C02FK1484/p1677765605320289). Looked again this morning and they only include the SDK as a dev dependency. I re-added it.

vendor/gems/ruby_spec_helpers/lib/ruby_spec_helpers/capybara_configuration.rb
samandmoore: yea the code you removed seemed to do the opposite of the warning, but if the warning is gone and the code works, great for now.

samandmoore (approved):
domainlgtm

smudge (approved):
platform LGTM as well!

PR #184, opened 2023-03-03
feat: replace vendored style_closet w/ Primer

HipsterBrown:
### Summary

Fixes https://github.com/orgs/Betterment/projects/10?pane=issue&itemId=21046385

This PR is an exploration to replace the vendored style_closet gem with a documented, out of the box design system that integrates well with Rails. 

### Other Information

As part of this exploration, stylesheet processing is handed off to [cssbundle-rails](https://github.com/rails/cssbundling-rails) using dart-sass, replacing the sass-rails gem. This provides the latest Scss syntax and ability to pull in styles from npm, i.e. `@primer/css` and `@primer/view-components`.

With the adoption of ViewComponents and Primer, there is further optimization available to create components for existing abstractions, like the Info Card, and replace simple_form with a more accessible alternative like [view_component-form](https://github.com/pantographe/view_component-form). 

I chose not to migrate the JS processing to jsbundling-rails since it was not required to adopt Primer or ViewComponent. 

If this feels like too much to do in a single PR, it _should_ be possible to take a more incremental approach by:

- adoption cssbundling-rails to process the existing stylesheets
- adding primer css and view components to the codebase
- migrating individual views / features over several PRs

This branch can serve as a blueprint for what the final result might look like.

Screenshots:
Screenshot 2023-03-03 at 16-51-43 Split Details test_split_enabled
Screenshot 2023-03-03 at 16-51-25 Split Details test_split_enabled
Screenshot 2023-03-03 at 16-51-16 Split Details test_split_enabled
Screenshot 2023-03-03 at 16-33-52 Test Track Admin
Screenshot 2023-03-03 at 16-33-41 Test Track Admin


.node-version
```diff
@@ -0,0 +1 @@
+14.19.1
```
HipsterBrown, line 1: This is matching our current version used by Rails apps.

Procfile.dev
```diff
@@ -0,0 +1,2 @@
+web: unset PORT && bin/rails server
+css: yarn build:css --watch
```
HipsterBrown, line 2: Added by the `rails css:install:sass` task

app/assets/stylesheets/typography/_base.scss
```diff
@@ -1,16 +1,91 @@
...
+  line-height: $lh-default;
+  min-width: 320px;
+
+  h1 {
+    @include h1-responsive;
+  }
+
+  h2 {
+    @include h2-responsive;
+  }
```
HipsterBrown, line 16: This is the closest equivalent of the `sc-responsive-typography` mixin from style_closet using Primer mixins.

app/controllers/admin/bulk_assignments_controller.rb
```diff
@@ -18,7 +18,7 @@ def enqueue_assignments
       flash[:success] = success_message
       redirect_to admin_split_path(@split)
     else
-      flash[:error] = "Please address the errors below."
+      flash[:danger] = "Please address the errors below."
```
HipsterBrown, line 21: Primer uses `:danger` rather than`:error` as the flash type for form errors.

bin/dev
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env sh
```
HipsterBrown, line 1: Created by cssbundling-rails install task.

app/controllers/admin/bulk_assignments_controller.rb
smudge: I remember there being some issue in other apps (maybe `clinic`?) where some framework code was emitting a flash message using a key that was incompatible with our frontend. Maybe it was `devise` or omniauth-related? Worth spot-checking that failure messages from some of the implicitly mounted rails engines will still show up properly

HipsterBrown: Rather than change these setters, I created a FlashComponent that will map the flash names to the `scheme` values expected by the Primer Banner component.

samandmoore (approved):
domainlgtm 

Love the flash component. 

This seems great to me. 

Anything we need to consider when deploying? It should just work with the asset pipeline based deploy steps we already have right?

Gemfile
```diff
@@ -87,3 +85,7 @@ group :test do
   gem 'timecop'
   gem 'db-query-matchers'
 end
+
+gem "primer_view_components", "~> 0.1.0"
```
samandmoore, line 89: Minor thing but can we put these two new gems up before the development and test stuff? Like where the other similar gems are.

app/assets/stylesheets/components/_inputs.scss
```diff
@@ -6,8 +6,8 @@
   @include icomoon;
   content: $icon-chevron-thin-down;
   font-size: 2.5rem;
-  bottom: 0.25em;
-  right: 0.5em;
+  bottom: $em-spacer-3;
+  right: $em-spacer-5;
```
samandmoore, line 10: What's the story around when to use the sass variables vs css variables?

samandmoore (changes requested):
Tafn on the platform stuff. Just one minor annoying request.

config/application.rb
```diff
@@ -28,6 +28,8 @@ class Application < Rails::Application
     # The default locale is :en and all translations from config/locales/*.rb,yml are auto loaded.
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     # config.i18n.default_locale = :de
+    require 'view_component'
+    require 'primer/view_components/engine'
```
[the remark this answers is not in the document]
HipsterBrown: > In config/application.rb, add after the application definition:

https://primer.style/view-components/#installation

I believe that is what the docs specify.

app/assets/stylesheets/components/_inputs.scss
HipsterBrown: I went with whatever was documented: https://primer.style/css/support/variables
It seems like all the colors are currently CSS custom properties at the moment.

HipsterBrown:
Thanks for the feedback! Related to the deployment question:

> When you deploy your application to production, the css:build task attaches to the assets:precompile task to ensure that all your package dependencies from package.json have been installed via yarn, and then runs yarn build:css to process your stylesheet entrypoint, as it would in development. This output is then picked up by the asset pipeline, digested, and copied into public/assets, as any other asset pipeline file.

https://github.com/rails/cssbundling-rails#css-bundling-for-rails

It should "just work" 🤞

config/application.rb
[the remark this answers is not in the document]
smudge: Do they mean _after_ as in after the `end`, or after the `do`?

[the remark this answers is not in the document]
smudge: (As in, I think they might mean at the bottom of the file.)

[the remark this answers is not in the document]
HipsterBrown: Now I'm not sure because the demo app in their repo place these requires after their railties. 😆 

https://github.com/primer/view_components/blob/main/demo/config/application.rb#L9-L11

I'll move them to match the demo app and confirm that works.

samandmoore (approved):
domainlgtm platformlgtm

smudge (approved):
domain LGTM!

PR #185, opened 2023-03-20
fix: unblock deploys (upgrade ddtrace & config, switch to terser, and enable spec coverage)

smudge:
### Summary

Older versions of the `ddtrace` gem are now doing some in-place upgrade magic that breaks the `datadog.rb` initializer as it was written. This change:

- Upgrades to the latest version of ddtrace
- Fixes the initializer to work on that version
- Loads `ddtrace` (but disables tracing) in tests so that we can detect such breakages going forward

For the sake of unblocking deployments, this also:
- Switches from `:uglifier` to `:terser` for ES6 support
- Enables `:terser` js compression in CI so that we can be alerted to build failures sooner

/no-platform


HipsterBrown (approved):
DomainLGTM

PR #187, opened 2023-03-28
Test against a matrix of maintained rubies

smudge:
### Summary

The idea is to always be testing against this whole list:
https://www.ruby-lang.org/en/downloads/branches/

### Other Information

Not sure if 3.1 and 3.2 tests will pass -- I'm still waiting for them to install locally, and figured I'd get a PR up in the meantime.

/no-platform


.github/workflows/ci.yml
```diff
@@ -13,6 +13,9 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ruby: ['3.0', '3.1']
```
smudge, line 18: Rails 6.0 does not support Ruby 3.2 -- and actually only 7.0.X officially supports 3.2 at this point. We can update the matrix when we get onto newer Rails.

samandmoore (approved):
domainlgtm

PR #189, opened 2023-04-26
Support updating owner via migration

jesseproudman:
### Summary

Supports updating ownership of a split from testtrack-cli 1.5.0.

### DIFF
--- a/app/models/split_upsert.rb
+++ b/app/models/split_upsert.rb
@@ -1,13 +1,13 @@
 class SplitUpsert
   include ActiveModel::Model
 
-  attr_accessor :app, :name, :decided_at, :finished_at, :require_app_name_prefix
+  attr_accessor :app, :name, :owner, :decided_at, :finished_at, :require_app_name_prefix
   attr_reader :weighting_registry
 
   validate :split_must_be_valid
 
   def save # rubocop:disable Metrics/AbcSize
-    split.assign_attributes(finished_at: finished_at, decided_at: decided_at, require_app_name_prefix: require_app_name_prefix)
+    split.assign_attributes(finished_at: finished_at, decided_at: decided_at, require_app_name_prefix: require_app_name_prefix, owner: owner)
     split.reassign_weight(merged_registry) unless split.registry == merged_registry
     return false unless valid?
 
--- a/app/models/split_upsert.rb
+++ b/app/models/split_upsert.rb
@@ -7,7 +7,8 @@ class SplitUpsert
   validate :split_must_be_valid
 
   def save # rubocop:disable Metrics/AbcSize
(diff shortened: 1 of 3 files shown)


spec/controllers/api/v2/migrations/splits_controller_spec.rb
```diff
@@ -19,12 +19,13 @@
...
     describe '#create' do
       it "creates a new split with desired weightings" do
-        post :create, params: { name: 'default_app.foobar', weighting_registry: { foo: 10, bar: 90 } }
+        post :create, params: { name: 'default_app.foobar', owner: 'b4c', weighting_registry: { foo: 10, bar: 90 } }
 
         expect(response).to have_http_status(:no_content)
         split = Split.find_by(name: 'default_app.foobar', owner_app: default_app)
         expect(split).to be_truthy
         expect(split.registry).to eq 'foo' => 10, 'bar' => 90
+        expect(split.owner).to eq 'b4c'
```
smudge, line 28: Nitpicking, but no need to use internal team names on a public repo 😉 

```suggestion
        post :create, params: { name: 'default_app.foobar', owner: 'team-name', weighting_registry: { foo: 10, bar: 90 } }

        expect(response).to have_http_status(:no_content)
        split = Split.find_by(name: 'default_app.foobar', owner_app: default_app)
        expect(split).to be_truthy
        expect(split.registry).to eq 'foo' => 10, 'bar' => 90
        expect(split.owner).to eq 'team-name'
```

app/models/split_upsert.rb
```diff
@@ -1,13 +1,13 @@
...
 
-  attr_accessor :app, :name, :decided_at, :finished_at, :require_app_name_prefix
+  attr_accessor :app, :name, :owner, :decided_at, :finished_at, :require_app_name_prefix
   attr_reader :weighting_registry
 
   validate :split_must_be_valid
 
   def save # rubocop:disable Metrics/AbcSize
-    split.assign_attributes(finished_at: finished_at, decided_at: decided_at, require_app_name_prefix: require_app_name_prefix)
+    split.assign_attributes(finished_at: finished_at, decided_at: decided_at, require_app_name_prefix: require_app_name_prefix, owner: owner)
```
smudge, line 10: In the case where the client is not yet sending an `owner` param, do we want to do something to avoid overwriting an existing owner with `nil`?

smudge: (given that this is an "upsert" operation)

jesseproudman:
@smudge Specs now pass here.

smudge (approved):
domain LGTM && platform LGTM

jesseproudman:
I've got no merge rights here. :)

PR #192, opened 2023-05-02
Enable Rails 6.1 defaults

argvniyx-enroute:
### Summary
This has the results of running `rake app:update` and keeping obvious changes, as well as now loading Rails 6.1 defaults in `config/application.rb`. I followed https://lilyreile.medium.com/rails-6-1-new-framework-defaults-what-they-do-and-how-to-safely-uncomment-them-c546b70f0c5e to check how safe each default was. 

### Other Information


config/initializers/assets.rb
```diff
@@ -3,9 +3,12 @@
 # Version of your assets, change this if you want to expire all your assets.
 Rails.application.config.assets.version = '1.0'
 
-# Add additional assets to the asset load path
+# Add additional assets to the asset load path.
 # Rails.application.config.assets.paths << Emoji.images_path
+# Add Yarn node_modules folder to the asset load path.
+Rails.application.config.assets.paths << Rails.root.join('node_modules')
```
argvniyx-enroute, line 9: I'm not sure if this change is ok, happy to revert (or just comment it out) if this is wrong.

smudge: I think we should remove this (or comment it out) unless it's somehow necessary to get the Admin UI working -- are you able to boot up the app locally and get into the admin interface to test it?

argvniyx-enroute: Yup turns out this isn't needed, so I commented this out; just waiting on specs now

smudge (approved):
domain LGTM && platform LGTM -- thank you!

PR #193, opened 2023-05-15
Fix `parent_name` deprecation

argvniyx-enroute:
### Summary
`parent_name` is deprecated in Rails 6.1, so this PR changes a stray call into `module_parent_name`


smudge (approved):
domain LGTM && platform LGTM -- it would be nice to reproduce this, but for now I'm going to merge this to unblock the work you're doing

smudge (approved):
domain LGTM && platform LGTM -- it would be nice to reproduce this in the test suite, but for now I'm going to merge this to unblock the work you're doing

PR #194, opened 2023-05-23
pass id to quote to fix deprecation

argvniyx-enroute:
### Summary
Noticed the following deprecation warning while running specs:

```
DEPRECATION WARNING: Passing an Active Record object to `quote` directly is deprecated
and will be no longer quoted as id value in Rails 7.0.
 (called from update_assignments at /Users/ricardogarza/src/test_track/app/models/bulk_reassignment.rb:66)
```

This PR simply passes an id to `quote` directly, to avoid the warning.


samandmoore (approved):
domainlgtm platformlgtm

PR #199, opened 2023-06-30
chore: use newest rails version to address dependabot alert

samandmoore:
fixes https://github.com/Betterment/test_track/security/dependabot/54


effron (approved):
domainLGTM && platformLGTM!

PR #202, opened 2023-09-28
Ruby 3.2.2 upgrade

argvniyx-enroute:
### Summary
/task https://app.asana.com/0/0/1205605778894636/f

Update `test-track` to use ruby 3.2.2

- Bump `.ruby-version`
- Update code to be ruby 3.2.2 compatible
- Clean up loose ends from Rails upgrades (deprecations, schema updates)

I had to bump the `nanasess/setup-chromedriver` action step to use v. 117, because 115 was failing (though it was a bit of a cargo cult so do let me know if there's a better way to go about this)
image

### Other Information


spec/system/admin_variant_retirement_create_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.describe 'admin can retire a variant for a split' do
   let(:split_page) { app.admin_split_show_page }
-  let(:variant_page) { app.admin_variant_details_edit_page }
+  let!(:variant_page) { app.admin_variant_details_edit_page }
```
argvniyx-enroute, line 5: part of a dummy commit, since actions failed for the second to last commit due to docker

smudge: If you do `git commit --amend` (to rewrite the last commit) and then `get push -f` (to force push to the branch) I think you can accomplish something similar. But GH actions should also be re-runnable through the GH UI.

argvniyx-enroute: duh, didn't think of using `--amend` 🙃 (I did think of retrying from the UI, but couldn't find the button to do so? and figured it's maybe because I forked this repo (but it could just be me missing it))

argvniyx-enroute: Ok, undid this (and had the chance to try out `--amend` in the process)

.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: ['3.0', '3.1']
+        ruby: ['3.0', '3.1', '3.2.2']
```
smudge, line 18:
```suggestion
        ruby: ['3.0', '3.1', '3.2']
```

Minor, but we commonly don't care too much about patch versions when it comes to the public repo test configs.

smudge (dismissed):
domain LGTM

smudge (changes requested):
platform TAFN -- just one suggestion

smudge (approved):
domain LGTM && platform LGTM

smudge:
Looks like merging is blocked -- filed an issue since I think this is an error with the branch configs.

smudge:
@argvniyx-enroute merging is unblocked now, so I'm merging this!

PR #212, opened 2024-03-20
Add Optional Environment Banner

qcannon:
### Summary

Currently, if an organization uses Test Track for multiple environments (e.g. production and stage), it can be a bit difficult to discern at first glance for which environment a user is managing a given split test.

This proposed change gives users the option to set a `DEPLOYMENT_ENV_LABEL` variable in the `.env` or `.env.local` file, which will subsequently be rendered in a bright red banner, like so:

Screenshot 2024-03-20 at 5 23 52 PM

The banner is intentionally a tad obnoxious to make it glaringly obvious to the user which environment they are in. 

When no `DEPLOYMENT_ENV_LABEL` variable is set, the header renders as usual:

Screenshot 2024-03-20 at 5 23 24 PM

I envision organizations using the existing state for production and the banner state for a non-production environment (or perhaps vice-versa). 

### DIFF
--- a/spec/requests/admin/split_details_spec.rb
+++ b/spec/requests/admin/split_details_spec.rb
@@ -7,17 +7,19 @@ RSpec.describe Admin::SplitsController do
     let(:default_app) { FactoryBot.create(:app, name: "default_app", auth_secret: "6Sd6T7T6Q8hKcoo0t8CTzV0IdN1EEHqXB2Ig4raZsOf") }
     let(:admin) { FactoryBot.create(:admin) }
     let(:deployment_env_label) { 'Stage Environment' }
+    let(:banner_tag) { "<p class=\"banner-text\">" }
 
     before do
       login_as admin
     end
 
     context 'when no DEPLOYMENT_ENV_LABEL is set' do
-      it 'renders without banner' do
+      it 'renders without environment label banner' do
         get '/admin'
 
         expect(response).to have_http_status :ok
 
+        expect(response.body).not_to include(banner_tag)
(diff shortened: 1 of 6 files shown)


smudge (changes requested):
Love it! 👏 

Any chance you can add a test for this? We have [this `with_env`](https://github.com/Betterment/test_track/blob/555c0d6a831c9395c7e3e03381c473426894c752/spec/support/environment_spec_helper.rb#L2) helper that allows us to change specific ENV vars on a per-example basis.

qcannon:
@smudge oh yes good call, and thank you for the tip on `with_env` helper! Pushed that change in https://github.com/Betterment/test_track/pull/212/commits/44389995ac4fce4bd6271f5a4db54e06c3482d95

spec/requests/admin/split_details_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'rails_helper'
+
+RSpec.describe Admin::SplitsController do
+  include Warden::Test::Helpers
+
+  describe 'GET /admin' do
+    let(:default_app) { FactoryBot.create(:app, name: "default_app", auth_secret: "6Sd6T7T6Q8hKcoo0t8CTzV0IdN1EEHqXB2Ig4raZsOf") }
```
smudge: Do the tests pass without the `auth_secret` value passed? I'm wondering if this (fake, definitely fake! lol) auth secret is only necessary for API-based tests.

The `login_as` below should do the real auth for these requests.

smudge (approved):
domain LGTM && platform LGTM!

PR #234, opened 2024-11-14
Split "Experiment Details" out from feature gate info, add more experiment metadata

smudge:
### Summary

This is the foundation for what may become a set of PRs that improve the experience for managing metadata for split _experiments_ (& better distinguishing them from _feature gates_ by way of a new UI section):

Screenshot 2024-11-14 at 5 25 56 PM

This section only appears for non-feature-gate splits, and enables users to add & manage metadata about the experiment.

**None of this data does anything!** It's merely intended as a convenience for users of TestTrack who wish to perform downstream analysis of splits, and want to have metadata colocated with the rest of their split data:

Screenshot 2024-11-14 at 5 28 31 PM

### Bonus

This also adds takeaways to the split decision view, so that takeaways can be captured during the decision event:

takeaways

### Other Information

There are other changes I'd like to make to this (like, more thoughtful validations, and potentially more configurability) but this PR is meant to set the foundation.

This also ships with completely empty configs for `ExperimentDetail.available_tests` `ExperimentDetail.available_segments` -- these can be configured at boot time by dropping a file in `config/initializers` that populates these hashes (`name => description`). Description is not currently used but I'd like to put it in a tooltip (future PR).

### DIFF
--- a/db/migrate/20241113191054_add_experiment_details.rb
+++ b/db/migrate/20241113191054_add_experiment_details.rb
@@ -12,7 +12,7 @@ class AddExperimentDetails < ActiveRecord::Migration[7.0]
   end
 
   def change
-    create_table :experiment_details, id: :uuid do |t|
+    create_table :experiment_details, id: :uuid, default: -> { 'uuid_generate_v4()' } do |t|
       t.references :split, null: false, foreign_key: true, type: :uuid
       t.string :control_variant
       t.date :start_date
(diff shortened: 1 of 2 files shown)


app/models/experiment_detail.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def end_date
+    super || split.decided_at&.to_date
+  end
+
+  def control_variant
+    super || split.registry.keys.find { |v| v == 'control' }
+  end
+
+  def self.available_tests
```
jonmauney, line 18: where will we actually enumerate these? (same with `available_segments`)

jonmauney (changes requested):
couple notes!

PR #238, opened 2025-01-30
feat: Upgrade to Rails 7.1

argvniyx-enroute:
### Summary

This bumps Rails to 7.1 and addresses deprecation warnings, with an exception. It also drops Ruby 3.1 testing and bumps rubocop's `TargetRubyVersion` to 3.2 (all resulting lints were `rubocop -a`'d, since they were `Style/HashSyntax`)

### Other Information
There's a couple of deprecation warnings left to deal with, namely:

```
DEPRECATION WARNING: `Rails.application.secrets` is deprecated in favor of `Rails.application.credentials` 
and will be removed in Rails 7.2.
```
and

```
DEPRECATION WARNING: Your `secret_key_base` is configured in `Rails.application.secrets`, 
which is deprecated in favor of `Rails.application.credentials` and will be removed in Rails 7.2.
```

These both come from `devise`, which has a `SecretKeyFinder`: https://github.com/heartcombo/devise/blob/v4.9.4/lib/devise/secret_key_finder.rb. That'll check if it can find the secret key in the different possible configs (`secrets` and `credentials`) and call the appropriate method. Since we're using `secrets`, we'll trigger the deprecation warning.

This app doesn't call `secrets` explicitly itself, and we rely (AFAICT) on setting the `SECRET_KEY_BASE` env var. Since `Rails::Application#secret_key_base` ultimately looks for that, I think we can at least punt on the secrets -> credentials migration.


config/initializers/devise.rb
```diff
@@ -238,7 +238,7 @@
   # config.omniauth :github, 'APP_ID', 'APP_SECRET', scope: 'user,public_repo'
   config.omniauth :saml,
     issuer: ENV.fetch('SAML_ISSUER', nil),
-    idp_sso_target_url: ENV.fetch('IDP_SSO_TARGET_URL', nil),
+    idp_sso_service_url: ENV.fetch('IDP_SSO_TARGET_URL', nil),
```
argvniyx-enroute, line 241: We're currently using `ruby-saml` 1.17.0, and apparently this changed to `idp_sso_service_url` since 1.12.0 (https://github.com/SAML-Toolkits/ruby-saml/blob/c9685a98a9c00c38af65364e70a7315eb49a2222/UPGRADING.md?plain=1#L16-L23).

Not sure if devise does something to account for this, but I did [see an error in datadog that was similar to what prompted Rob to update authenticated_persona](https://app.datadoghq.com/logs?query=service%3Atest-track-app%20status%3Aerror%20Omniauth&agg_m=count&agg_m_source=base&agg_t=count&cols=core_host%2Ccore_service%2Clog_name&event=AwAAAZSydFubEwXTFgAAABhBWlN6aHhBb0FBQUJla0swaURIVzF3Q0EAAAAkMDE5NGIyNzgtMGUzZi00YjE1LWExNjEtMWQxMWJiZmZmY2U1AAJV6A&fromUser=true&graphType=flamegraph&messageDisplay=inline&refresh_mode=sliding&sort=time&spanID=3494412893000385356&storage=hot&stream_sort=desc&viz=stream&from_ts=1729776983724&to_ts=1730381783724&live=true)

config/environments/test.rb
```diff
@@ -29,7 +29,7 @@
   config.cache_store = :null_store
 
   # Raise exceptions instead of rendering exception templates.
-  config.action_dispatch.show_exceptions = false
+  config.action_dispatch.show_exceptions = :none
```
argvniyx-enroute, line 32: Addresses

```
DEPRECATION WARNING: Setting action_dispatch.show_exceptions to false is deprecated. 
Set to :none instead.`
```

.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: ['3.1', '3.2']
```
argvniyx-enroute, line 18: So, it turns out we don't have a `with_transactional_lock` version that is both able to support Rails 7.1 and Ruby <3.2; not because it's not a vallid combo, but because we just happened to drop Ruby <3.2 support before adding Rails 7.1 support 🙃 

I figured that, before trying anything else, I'd call this out here. I feel that this is _ok_ish since it's not, say, a gem, but lmk if that's not the case.

argvniyx-enroute:
just in case: I don't have permissions to merge this cc @smudge

argvniyx-enroute:
@smudge alright! SRE figured out that the lingering workflow was a stray branch protection rule, this should be good to go!

PR #249, opened 2025-04-10
Add admin page to show current test track assignments

stevenjackson:
### Summary

For longer running experiments with a specific audience in my mind (like preferred beta testers) it's useful to use the bulk assignments capability to enable those features. However if there are mulitple instances of "adding users to the beta", one doesn't have visibility into what assignments may already be present.  This view is intended to give admins more confidence that the intended assignments are in place.

### Other Information

I've arbitrarily selected a limit of 1000 assignments to avoid performance issues for splits that have massive bulk assignments of huge populations.  This is entirely too many records to skim IMO, but I didn't want to take on pagination as part of this minor quality of life improvement.

This is checked via a system spec, but creating those assignments results in at least 5000 inserts and adds about 5 seconds to the test.  I debated testing with passing the limit as an optional query parameter, but wasn't sure that would give the necessary confidence that TT wouldn't allow a user to exceed the limit.  But maybe that's ok - allowing `?limit=any` while defaulting to a small number of records would allow for a speedier test and allow for use cases with larger data sets.

**Note**: Rendering the page locally is quite quick (median < 200ms).

### Screenshots

Screenshot 2025-04-10 at 4 21 29 PM
Screenshot 2025-04-10 at 4 10 20 PM


spec/system/admin_split_assignments_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'rails_helper'
+
+RSpec.describe 'bulk assign flow' do
```
ivangreene, line 3: This description might be copied? Should it be more like "view assignments page"

app/views/admin/split_assignments/index.html.erb
```diff
@@ -0,0 +1,27 @@
...
+      </thead>
+      <tbody>
+        <% @assignments.group_by(&:variant).each do |variant, assignments| %>
+          <% assignments.map(&:visitor).flat_map(&:identifiers).group_by { |identifier| identifier.identifier_type }.each do |type, identifiers| %>
+            <% identifiers.each do |identifier| %>
+              <tr>
+              <td ><%= type.name%></td>
+              <td ><%= identifier.value%></td>
+              <td><%= variant %></td>
+              </tr>
```
ivangreene, line 20:
```suggestion
              <tr>
                <td><%= type.name%></td>
                <td><%= identifier.value%></td>
                <td><%= variant %></td>
              </tr>
```
Small formatting things here

app/controllers/admin/split_assignments_controller.rb
```diff
@@ -0,0 +1,10 @@
+class Admin::SplitAssignmentsController < AuthenticatedAdminController
+  VIEW_LIMIT = 1000
+
+  def index
+    split = Split.find(params.require(:split_id))
+    @split_name = split.name
+    @assignments = split.assignments.includes(visitor: { identifiers: :identifier_type }).order(updated_at: :desc).limit(VIEW_LIMIT)
```
ivangreene, line 7: Someone might have a stronger opinion on whether an index is necessary here. I think it's optional, because we want to preserve insert/update efficiency, and this query is not in the critical path (it is loaded by users on demand)

spec/system/admin_split_assignments_spec.rb
stevenjackson: Thank you! Updated

app/controllers/admin/split_assignments_controller.rb
jmileham: One thing that might make this slow on big assignment datasets is having an order by clause, which requires fetching the whole assignment set even in light of the limit clause. We could consider removing it. My guess is it'd make this very fast regardless of assignment count.

stevenjackson: I think that's reasonable. My only hesitation is our experience has been the assignments you care about most are the ones you manually assigned last week.  But I'm already leaning YAGNI - either the data set is skimmable or you need search, there's not a lot of value in the in-between states.

jmileham: Yeah, if we see this as a scrappy solution for low-scale assignments it makes sense, and we can iterate if we see value.

jmileham: Sorting in ruby after retrieval might be a nice feature for skimmability under 1k rows, though it might mislead people to think that absence of a sorted row means something.

.rubocop_todo.yml
```diff
@@ -11,11 +11,12 @@ Betterment/NonStandardActions:
   Exclude:
     - 'config/routes.rb'
 
-# Offense count: 13
+# Offense count: 14
```
jmileham, line 14: Note to self I think this cop can be configured with models that aren't considered "owned" which we should use to knock this exclusion list out.

jmileham (approved):
Domain Lgtm platform lgtm!

samandmoore:
You said you didn't want to take on pagination, but does that have to be much harder than pulling in kaminari? 

I'm fine with landing this and adding pagination separately if the concern is mostly just a big ol' diff. 

Just feels oddly limiting to only show the first 1K with no way to see the rest. 

This also makes me think we should build a "view visitor" page where you can punch in an identifier and see the details for that visitor.

stevenjackson:
@samandmoore 

I'm mostly not sure pagination makes sense either.  If I have 570K overrides, how many pages would I click through?  Agree the arbitrary limit feels strange.  My other thought was perhaps allowing overriding limit via url param - if the admin really wants to comb through a 500K table, getting it in one view may be  preferred.  Then `1000` ends up being the "reasonable default" vs an unsurmountable blocker.

I like the idea of a visitor page!  I think that gets much closer to what I would be curious about.  

I see use cases like:
* Did the overrides I just assign work?
* Are we segmenting the population as expected?
* Is this user seeing weird behavior because of some unexpected combination of flags?
* Is this flag being used anymore?  Do the exceptions still make sense?

samandmoore:
Yea that tracks too. That's exactly what lead me to suggesting a way to look up an individual visitor. 

Well, if you find yourself working on that, I will def give it a review 🙂

Thanks for improving the admin experience!

PR #256, opened 2025-06-03
feat: Run `rake app:update` and opt into Rails 7.1 defaults

argvniyx-enroute:
### Summary

> Provide a general description of the code changes in your pull
request... were there any bugs you had fixed? If so, mention them. If
these bugs have open GitHub issues, be sure to tag them here as well,
to keep the conversation linked together.

I ran `bundle exec rake app:update` and picked changes I thought were sensible (so, benign changes like quoting/prose changes, or new defaults added by the 7.1 templates)

I'm also updating `load_defaults` to 7.1 (since this doesn't depend on our internal runtime configuration). We tested this on our stage instance.

### Other Information

I had to bump `primer_view_components` to 0.14.0, since that's the first version that adds Rails 7.1 support (see: https://github.com/primer/view_components/issues/2318)

/domain @JELaVallee
/domain @Irving-Betterment 
/platform @smudge

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,7 +21,7 @@ gem 'turbolinks'
 gem 'jbuilder', '~> 2.0'
 
 gem "cssbundling-rails", "~> 1.1"
-gem "primer_view_components", "~> 0.1.0"
+gem "primer_view_components", "~> 0.14.0"
 
 gem 'puma', '~> 5.6'
 gem 'nokogiri'
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,8 +1,10 @@
+require "active_support/core_ext/integer/time"
+
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
   # Code is not reloaded between requests.
-  config.cache_classes = true
+  config.enable_reloading = false
 
   # Eager load code on boot. This eager loads most of Rails and
   # your application in memory, allowing both threaded web servers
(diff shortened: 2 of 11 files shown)


Gemfile
```diff
@@ -21,7 +21,7 @@ gem 'turbolinks'
 gem 'jbuilder', '~> 2.0'
 
 gem "cssbundling-rails", "~> 1.1"
-gem "primer_view_components", "~> 0.1.0"
+gem "primer_view_components", "~> 0.14.0"
```
smudge, line 24: We have no visual regression testing for this app -- did you get a chance to boot it up locally and check the admin panel for anything obviously broken?

config/environments/production.rb
```diff
@@ -1,8 +1,10 @@
+require "active_support/core_ext/integer/time"
+
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
   # Code is not reloaded between requests.
-  config.cache_classes = true
+  config.enable_reloading = false
```
smudge, line 7: We still have `cache_classes =` in some of the other environment files -- is this largely just a config rename?

config/environments/production.rb
```diff
@@ -61,20 +77,17 @@
...
-  config.active_support.deprecation = :notify
-
-  # Use default logging formatter so that PID and timestamp are not suppressed.
-  config.log_formatter = Logger::Formatter.new
-
-  # Use a different logger for distributed setups.
-  # require "syslog/logger"
-  # config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new "app-name")
+  # Don't log any deprecations.
+  config.active_support.report_deprecations = false
```
smudge, line 81: Does this change anything about `config.active_support.deprecation = :notify`?

config/environments/production.rb
argvniyx-enroute: Yes. `cache_classes` is still kept around for backwards compatibility, but it'll eventually be dropped in favor of this.

config/environments/production.rb
argvniyx-enroute: It does! This'll make the default behavior become `:silence`. The reasoning behind jumping onto the new config is that we don't actually subscribe to these events.

Gemfile
JELaVallee: Yup! Didn't set the PR to "ready" until we gave it a tire-kick in stage... splits are loading fine on the summary and detail view and I was able to add a user to a split manually w/o error in the controller action or background job. 😄

config/environments/production.rb
JELaVallee: inquiry: Should they be getting subscribed-to/logged?

config/environments/production.rb
argvniyx-enroute: We can, yeah. Really the only reason I left other files untouched was that they had additional logic; I figured it could read a bit weird if we updated the config whilst relying on an envvar with a different name.

smudge: I don't mind -- I think I'd rather make the naming inconsistency more explicit and be consistent about which config we use.

argvniyx-enroute: Done! A test seems to have failed due to a timeout, but I don't think it's related to this.

smudge (approved):
domain LGTM && platform LGTM

argvniyx-enroute:
Hmm, still stuck on the test job 🤔 @smudge would you mind re-running the workflow again, just to rule out these failures being caused by my changes?

PR #282, opened 2026-04-08
Update cssbundling-rails to 1.4

chtakahashi:
v1.4 enables use of pnpm


smudge (approved):
domain LGTM && platform LGTM

PR #283, opened 2026-05-04
fix: ensure individually_overridden is reliably cleared

jmileham:
individually_overridden was not being maintained as a trustworthy signal for "this assignment is currently a manual per-visitor override." Two related bugs caused it to persist stale true values after bulk operations:

1. BulkReassignment's SQL UPDATE never included individually_overridden in its SET clause, leaving the flag true after bulk reassignment.

2. ArbitraryAssignmentCreation#individual_override? had a stickiness term (assignment.individually_overridden || !bulk_assignment?) that carried the flag forward onto the result even when a bulk operation superseded an individually-overridden record via the ArbitraryAssignmentCreation path. With the BulkReassignment bug fixed, this stickiness has no legitimate purpose — the flag will now be correctly false on any record a bulk operation has touched, so the term simplifies to !bulk_assignment?.

A backfill migration corrects existing records where individually_overridden is true but context shows the last operation was not an individual override. The history of individual overrides is preserved in previous_assignments, though those records may themselves be tainted by these bugs.

@samandmoore this is some prework buggy data identified before I take a stab at making stale individual overrides time out.

@jonmauney is there any possible impact on downstream processing? I don't think this value (which had to my eye buggy values after some bulk operations) is exposed on any API or used in filtering, and I believe all A/B test analysis is performed based on client-generated events now, so this feels low risk, but let me know if anything feels like it might be depending on this!

### DIFF
--- a/db/migrate/20260504164518_clear_stale_individually_overridden_flags.rb
+++ b/db/migrate/20260504164518_clear_stale_individually_overridden_flags.rb
@@ -1,0 +1,14 @@
+class ClearStaleIndividuallyOverriddenFlags < ActiveRecord::Migration[7.2]
+  def up
+    execute <<~SQL
+      UPDATE assignments
+      SET individually_overridden = false
+      WHERE individually_overridden = true
+        AND context IS DISTINCT FROM 'individually_overridden'
+    SQL
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
--- a/spec/models/arbitrary_assignment_creation_spec.rb
+++ b/spec/models/arbitrary_assignment_creation_spec.rb
@@ -233,15 +233,15 @@ RSpec.describe ArbitraryAssignmentCreation do
         end
       end
 
-      it "remains during a bulk assignment" do
+      it "is cleared during a bulk assignment" do
         existing_assignment = FactoryBot.create(:assignment, existing_assignment_params.merge(individually_overridden: true))
         bulk_assignment = FactoryBot.create(:bulk_assignment, split: existing_assignment_params[:split])
 
         ArbitraryAssignmentCreation.create!(params.merge(bulk_assignment_id: bulk_assignment.id))
 
         expect(existing_assignment.reload.bulk_assignment).to eq bulk_assignment
-        expect(existing_assignment).to be_individually_overridden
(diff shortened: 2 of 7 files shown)


spec/models/arbitrary_assignment_creation_spec.rb
```diff
@@ -233,15 +233,15 @@
         end
       end
 
-      it "remains during a bulk assignment" do
```
jmileham, line 236: This seemed like "testing behavior that isn't part of the requirements" when I paged through the archived repository. I couldn't find any documented reason that this would be desirable behavior. It basically committed to what I think is a bug here.

db/migrate/20260504164518_clear_stale_individually_overridden_flags.rb
```diff
@@ -0,0 +1,14 @@
+class ClearStaleIndividuallyOverriddenFlags < ActiveRecord::Migration[7.2]
+  def up
+    execute <<~SQL
+      UPDATE assignments
+      SET individually_overridden = false
+      WHERE individually_overridden = true
+        AND context IS DISTINCT FROM 'individually_overridden'
```
jmileham, line 7: I had an ephemeral test to validate that this migration behaved as expected, but it felt exotic to test a migration so I deleted it.

jonmauney:
@jmileham we're leaning on what I think is a server-side fired "splitAssigned" event ([here](https://github.com/Betterment/retail/blob/main/%40betterment/retail/index.ts#L50)), and it doesn't care about `individually_overridden` directly.  That said, the old split reports code _does_ [automatically ignore individually_overriden as a context](https://github.com/Betterment/research/blob/main/split_reports/helpers/split_config_helpers.R#L75), which I assume is sourced from this boolean, but 1) this has been sunset and 2) i think this was a niche enough TT feature to where it probably didn't affect outcomes of tests. Beyond that, no usage of that column in [ETL](https://github.com/Betterment/etl/blob/main/dbt/models/staging/test_track/stg_test_track__assignments.sql).

jmileham:
Thanks @jonmauney - I think I've found another data quality problem here where this was only ever set on supersession of a previous explicit assignment as well which also made it non-complete for "first time override" case, so I'm gonna add to this PR for that case as well, and backfill this column to a complete set of actual individual overrides

